PR #115, opened 2014-11-09
EmailAddressCollection.contains overloaded to check email address

AmirHossein:
Default `EmailAddressCollection.contains` uses `java.util.List.contains` which doesn't work as excepted in usual cases (finding and address in collection). Overloaded method only check email address.

**The Problem**
I have multiple aliases and need to filter search results to  specific address. I tried search filter like below but it absolutely doesn't work:
```
String myAddress = "xxx@yyy.zz";
SearchFilter.SearchFilterCollection collection = new SearchFilter.SearchFilterCollection(LogicalOperator.Or);
collection.add(new SearchFilter.ContainsSubstring(EmailMessageSchema.ToRecipients, myAddress));
collection.add(new SearchFilter.ContainsSubstring(EmailMessageSchema.CcRecipients, myAddress));
collection.add(new SearchFilter.ContainsSubstring(EmailMessageSchema.BccRecipients, myAddress));
```
As i also have filter on an extended property, i can't use AQS too. So i load items and filter participants one by one using values of `EmailMessage.getToRecipients()`, `EmailMessage.getCcRecipients()` and `EmailMessage.getBccRecipients()` methods. These methods return `EmailAddressCollection` and i found out that  `EmailAddressCollection.contains()` uses `java.util.List.contains` and it checks exact object not address. So i add overloaded method to get and check String address.


vboctor:
I'm going to close this since it has been inactive for a while @AmirHossein, feel free to re-open once the feedback is addressed.

Issue #152, opened 2014-12-29
AutodiscoverService.tryGetEnabledEndpointsForHost(...) does not correctly close the request


evpaassen:
The `tryGetEnabledEndpointsForHost(...)` method does not correctly close the request when it throws a `ServiceLocalException` or returns early. The method should always close the request.

I will create a PR with a fix.

PR #153, opened 2014-12-29
Fix AutodiscoverService.tryGetEnabledEndpointsForHost(...) not closing the request properly

evpaassen:
Fixes OfficeDev/ews-java-api#152.

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/AutodiscoverService.java
+++ b/src/main/java/microsoft/exchange/webservices/data/AutodiscoverService.java
@@ -1458,91 +1458,89 @@ public final class AutodiscoverService extends ExchangeServiceBase implements
    * @throws Exception the exception
    */
   private boolean tryGetEnabledEndpointsForHost(String host,
-      OutParam<EnumSet<AutodiscoverEndpoints>> endpoints)
-      throws Exception {
+      OutParam<EnumSet<AutodiscoverEndpoints>> endpoints) throws Exception {
     this.traceMessage(TraceFlags.AutodiscoverConfiguration, String.format(
         "Determining which endpoints are enabled for host %s", host));
 
-    // We may get redirected to another host. And therefore need to limit
-    // the number
-    // of redirections we'll tolerate.
+    // We may get redirected to another host. And therefore need to limit the number of redirections we'll
+    // tolerate.
     for (int currentHop = 0; currentHop < AutodiscoverMaxRedirections; currentHop++) {
-      URI autoDiscoverUrl = new URI(String.format(
-          AutodiscoverLegacyHttpsUrl, host));
+      URI autoDiscoverUrl = new URI(String.format(AutodiscoverLegacyHttpsUrl, host));
 
       endpoints.setParam(EnumSet.of(AutodiscoverEndpoints.None));
 
-      HttpWebRequest request = new HttpClientWebRequest(this.getSimpleHttpConnectionManager());
+      HttpWebRequest request = null;
       try {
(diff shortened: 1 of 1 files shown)


src/main/java/microsoft/exchange/webservices/data/AutodiscoverService.java
```diff
@@ -327,50 +327,56 @@ private URI getRedirectUrl(String domainName) throws EWSHttpException,
     this.traceMessage(TraceFlags.AutodiscoverConfiguration, String.format(
```
evpaassen: What do you mean with tracing exceptions? In this library, tracing is all about requests and responses, which are still being traced when exceptions are thrown. Do you mean the missing error logging on exceptions perhaps? I think that when the exceptions are bubbling up, we don't have to log them here.

src/main/java/microsoft/exchange/webservices/data/AutodiscoverService.java
```diff
@@ -1458,91 +1464,89 @@ else if (this.getCredentials()instanceof X509CertificateCredentials)
...
+    // tolerate.
     for (int currentHop = 0; currentHop < AutodiscoverMaxRedirections; currentHop++) {
-      URI autoDiscoverUrl = new URI(String.format(
-          AutodiscoverLegacyHttpsUrl, host));
+      URI autoDiscoverUrl = new URI(String.format(AutodiscoverLegacyHttpsUrl, host));
 
       endpoints.setParam(EnumSet.of(AutodiscoverEndpoints.None));
 
-      HttpWebRequest request = new HttpClientWebRequest(this.getSimpleHttpConnectionManager());
+      HttpWebRequest request = null;
```
evpaassen: I'm not very comfortable with refactoring these methods myself. Perhaps we open an issue or address it in a separate PR.

evpaassen:
I've added two commits which should address the remaining connection leaks and the unlogged exceptions.

src/main/java/microsoft/exchange/webservices/data/AutodiscoverService.java
```diff
@@ -318,64 +323,64 @@ private void writeLegacyAutodiscoverRequest(String emailAddress,
...
+        serviceCredentials.preAuthenticate();
+
+        // Apply credentials to the request
+        serviceCredentials.prepareWebRequest(request);
+      }
+
+      try {
+        request.prepareAsyncConnection();
+      } catch (Exception e) {
+        log.warn("Error while preparing asynchronous connection.", e);
```
evpaassen: I'm not sure what the severity should be here...

serious6: I dont think the logging is usefull in this case (since nobody will notice). We dont have some listening appenders either. So severity might be immaterial.

evpaassen: Libraries don't have appenders, but host applications do. Our application has a appender, so the log message will certainly be noticed if it has the right severity.

serious6: I would suggest 'error' in this case because the message underlines this. Maybe for documentation purpose we could add a section of 'how to declare logging and traces' to the readme file to also help others while logging with ews-java-api. Any comments?

serious6:
_merged with revision b7e8a656598f6d9cc6db6dc12d6e5720d76bc3ae_
@evpaassen, @MikeN123 **thanks for the contribution.**

PR #155, opened 2014-12-31
Warning cleanup.

barisaydinoglu:
- Warnings cleaned:
  - {c} is a raw type. References to generic type {c}<{t}> should be parameterized.
  - The serializable class {c} does not declare a static final serialVersionUID field of type long.
  - The static field {c}.{t} should be accessed in a static way.
  - Type safety: Unchecked cast from Object to {t}.
  - Unnecessary @SuppressWarnings("{w}").
- TimeSpanTest moved to **src/main/java** to **src/test/java**, and rewritten as JUnit test.

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/AbstractAsyncCallback.java
+++ b/src/main/java/microsoft/exchange/webservices/data/AbstractAsyncCallback.java
@@ -12,14 +12,14 @@ package microsoft.exchange.webservices.data;
 
 import java.util.concurrent.Future;
 
-abstract class AbstractAsyncCallback implements Runnable, Callback {
-  Future task;
+abstract class AbstractAsyncCallback implements Runnable, Callback<Object> {
+  Future<?> task;
   static boolean callbackProcessed = false;
 
   AbstractAsyncCallback() {
   }
 
-  AbstractAsyncCallback(Future t) {
+  AbstractAsyncCallback(Future<?> t) {
     this.task = t;
   }
 
--- a/src/main/java/microsoft/exchange/webservices/data/EwsUtilities.java
+++ b/src/main/java/microsoft/exchange/webservices/data/EwsUtilities.java
@@ -419,7 +419,6 @@ class EwsUtilities {
    * @return the t service object
    * @throws Exception the exception
    */
-  @SuppressWarnings("unchecked")
   protected static <TServiceObject extends ServiceObject>
   TServiceObject createEwsObjectFromXmlElementName(
       Class<?> itemClass, ExchangeService service, String xmlElementName)
@@ -518,8 +517,6 @@ class EwsUtilities {
    * @param items   the items
    * @return A TItem instance or null if no instance of TItem could be found.
    */
-
-  @SuppressWarnings("unchecked")
   static <TItem extends Item> TItem findFirstItemOfType(Class<TItem> cls,
       Iterable<Item> items) {
     for (Item item : items) {
@@ -679,7 +676,6 @@ class EwsUtilities {
    * @param value      the value
    * @param separators the separators
    */
-  @SuppressWarnings("unchecked")
   protected static <T extends Enum<?>> void parseEnumValueList(Class<T> c,
       List<T> list, String value, char... separators) {
     EwsUtilities.EwsAssert(c.isEnum(), "EwsUtilities.ParseEnumValueList",
@@ -740,7 +736,6 @@ class EwsUtilities {
    * @throws IllegalAccessException   the illegal access exception
    * @throws java.text.ParseException the parse exception
    */
-  @SuppressWarnings("unchecked")
   protected static <T> T parse(Class<T> cls, String value)
       throws InstantiationException, IllegalAccessException,
       ParseException {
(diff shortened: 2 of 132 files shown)


src/main/java/microsoft/exchange/webservices/data/IAsyncResult.java
```diff
@@ -16,7 +16,7 @@
  * Represents the stauts of Asynchronous operation.
  */
 
-public interface IAsyncResult extends Future {
+public interface IAsyncResult extends Future<Object> {
```
serious6: perhabs this is useful: [Java Generics - Unbounded Wildcards](http://docs.oracle.com/javase/tutorial/java/generics/unboundedWildcards.html)

src/main/java/microsoft/exchange/webservices/data/AbstractAsyncCallback.java
```diff
@@ -12,14 +12,14 @@
 
 import java.util.concurrent.Future;
 
-abstract class AbstractAsyncCallback implements Runnable, Callback {
-  Future task;
+abstract class AbstractAsyncCallback implements Runnable, Callback<Object> {
```
serious6: same issue: `Callback<Object>` -> `Callback<?>`

serious6:
perhabs we should have a closer look at the choosen generics as you mentioned by @vboctor´s line-comment. @barisaydinoglu I think some of the choosen generics are made to wide. for example:
```Java
public interface IAsyncResult extends Future<Object> {
  public Object getAsyncState();
  public WaitHandle getAsyncWaitHanle();
  public boolean getCompleteSynchronously();
  public boolean getIsCompleted();
}
```
I think it should be (using raw types)
```Java
public interface IAsyncResult extends Future<T> {
  public T getAsyncState();
  public WaitHandle getAsyncWaitHanle();
  public boolean getCompleteSynchronously();
  public boolean getIsCompleted();
}
```
perhabs this could be applied to more classes?

src/main/java/microsoft/exchange/webservices/data/IAsyncResult.java
vboctor: Thanks @serious6.  That is pretty much what I expected, Object matches everything except simple types.

vboctor:
As @serious6 mentioned in his comment with example about use of generics. I think we should use the following model:

- When accepting a parameter, we can use the most generic type that meets our needs (e.g. List<Object> or List<?>).
- When returning a type, or deriving from a type, I would rather use T (e.g. List<T>).  This enables code interacting with this type to get the full richness of T.

Now we can always use T to simplify the above (model mostly used in C#).  That will give the full power of generics with some extra generated code for each of the specific types used.

src/main/java/microsoft/exchange/webservices/data/EwsUtilities.java
```diff
@@ -1035,35 +1036,28 @@ protected static TimeSpan getXSDurationToTimeSpanValue(String xsDuration) {
...
-    //System.out.println(m.find());
     // Year
-    m = Pattern.compile("(\\d+)Y").matcher(xsDuration);
-    //System.out.println(m.find());
-    int year = 0;
-    if (m.find()) {
-      year = Integer.parseInt(m.group().substring(0,
-          m.group().indexOf("Y")));
-    }
+    //    m = Pattern.compile("(\\d+)Y").matcher(xsDuration);
```
barisaydinoglu: the parameter "year" is not used.

src/main/java/microsoft/exchange/webservices/data/EwsXmlReader.java
```diff
@@ -422,7 +422,7 @@ public String readValue() throws XMLStreamException,
     } else if (this.presentEvent.getEventType() == XmlNodeType.CHARACTERS
         && this.presentEvent.isCharacters()) {
                         /*
-			 * if(this.presentEvent.asCharacters().getData().equals("<")) {
```
barisaydinoglu: There are some errors with project formatter.
- Comment line formatting:
 - IntelliJ: does not format comment line indentation. So, every indentation of the code above is right based on your formatter.
 - Eclipse: formats (or changes) comment line indentation.
- google-styleguide for Java works different for Eclipse and Intellij.
- Mix of space and tab are being used as comment indentation character.
 - A tab could be a different number of columns depending on your environment, but a space is always one column.
 - In terms of how many spaces (or tabs) constitutes indentation, it's more important to be consistent throughout your code than to use any specific tab stop value.

serious6: @barisaydinoglu I also noticed this kind of gap in the styleguide and also realized that the provided formatters of the google-site have it´s differences. Running these against checkstyle also shows that there is something wrong with it and it is just bad that this topic keeps chasing us. Perhabs we can have a seperate PR with the things you noticed changed in the codeStyleSettings.xml. I opened #158 for this. Nevertheless the files with untouched code should be reverted in my opinion.

barisaydinoglu:
I've completed changes below:
- Reverted reindented lines which don't include any code changes.
- Cleaned ```@SuppressWarnings```.
- Inserted empty lines before fields.
- Corrected comments of ```serialVersionUID```.

serious6:
_merged with ref. 5aeb2e7aaddb89138959f02b1a1d86fa796c0405_
Thanks for the contribution

Issue #17, opened 2014-09-02
ExtendedPropertyDefinition parses Integers incorrectly


Noblebrown:
When ExtendedPropertyDefinition parses the SOAP response, it uses Integer.getInteger(attributeValue, 16) to parse the tag property, if present. This method does not parse numbers in an 0x0... hexadecimal format and will incorrectly parse all tags as 16. Integer.decode() should be used as this will read hexadecimal numbers.

PR #161, opened 2015-01-06
PropertyTag cast error fixed

barisaydinoglu:
When loading PropertyTag of ExtendedPropertyDefinition from xml, there was hex to integer cast error.

[```Integer.getInteger(attributeValue, 16)```](http://docs.oracle.com/javase/6/docs/api/java/lang/Integer.html#getInteger%28java.lang.String%2C%20int%29) line tries to determine the integer value of the system property with the specified name.
It fails and returns the second parameter as default value.

[```Integer.decode(attributeValue)```](http://docs.oracle.com/javase/6/docs/api/java/lang/Integer.html#decode%28java.lang.String%29) decodes a String into an Integer.
Accepts decimal, hexadecimal, and octal numbers.


azurecla:
Hi __@barisaydinoglu__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!

TTYL, AZPRBOT;

avromf:
This is a duplicate of issue #17. 
I've already put in a pull request in #137.

Issue #18, opened 2014-09-02
AttachmentCollection casts without a proper type check


Noblebrown:
In AttachmentCollection.validate(), there is a point at which attachments are cast to FileAttachment without a typecheck. This region is to check to see if the attachment is a contact photo in new attachments, but it is possible for an ItemAttachment to be processed. A type check should be implemented.

PR #163, opened 2015-01-07
Fix for AttachmentCollection validate

barisaydinoglu:
In ```AttachmentCollection.validate()```, type checking added to casting of ```Attachment``` to ```FileAttachment``` as fix for #18.

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/AttachmentCollection.java
+++ b/src/main/java/microsoft/exchange/webservices/data/AttachmentCollection.java
@@ -348,44 +348,37 @@ public final class AttachmentCollection extends
    */
   public void validate() throws Exception {
     // Validate all added attachments
-    boolean contactPhotoFound = false;
-    for (int attachmentIndex = 0; attachmentIndex < this.getAddedItems()
-        .size(); attachmentIndex++) {
-      Attachment attachment = this.getAddedItems().get(attachmentIndex);
-      if (attachment != null && attachment.isNew()) {
-
-        // At the server side, only the last attachment with
-        // IsContactPhoto is kept, all other IsContactPhoto
-        // attachments are removed. CreateAttachment will generate
-        // AttachmentId for each of such attachments (although
-        // only the last one is valid).
-        //
-        // With E14 SP2 CreateItemWithAttachment, such request will only
-        // return 1 AttachmentId; but the client
-        // expects to see all, so let us prevent such "invalid" request
-        // in the first place.
-        //
-        // The IsNew check is to still let CreateAttachmentRequest allow
-        // multiple IsContactPhoto attachments.
-        //
-        if (this.owner.isNew()
-            && this.owner.getService().getRequestedServerVersion()
-            .ordinal() >= ExchangeVersion.Exchange2010_SP2
-            .ordinal()) {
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@barisaydinoglu__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!

TTYL, AZPRBOT;

src/main/java/microsoft/exchange/webservices/data/AttachmentCollection.java
```diff
@@ -372,15 +372,17 @@ public void validate() throws Exception {
             && this.owner.getService().getRequestedServerVersion()
             .ordinal() >= ExchangeVersion.Exchange2010_SP2
             .ordinal()) {
-          FileAttachment fileAttachment = (FileAttachment) attachment;
+          if (attachment instanceof FileAttachment) {
```
serious6: perhabs these if`s should be combined for better overview?

src/main/java/microsoft/exchange/webservices/data/AttachmentCollection.java
```diff
@@ -372,15 +372,17 @@ public void validate() throws Exception {
...
+            FileAttachment fileAttachment = (FileAttachment) attachment;
 
-          if (fileAttachment.isContactPhoto()) {
-            if (contactPhotoFound) {
-              throw new ServiceValidationException(
-                  Strings.MultipleContactPhotosInAttachment);
-            }
+            if (fileAttachment.isContactPhoto()) {
+              if (contactPhotoFound) {
```
serious6: looking at the code I was wondering if this is a contradiction because `contactPhotoFound` is set to false and only changed to true if it is already true.

src/main/java/microsoft/exchange/webservices/data/AttachmentCollection.java
```diff
@@ -352,7 +352,7 @@ public void validate() throws Exception {
     for (int attachmentIndex = 0; attachmentIndex < this.getAddedItems()
         .size(); attachmentIndex++) {
       Attachment attachment = this.getAddedItems().get(attachmentIndex);
```
serious6: since the `Attachment` will not be changed again perhabs it should be declared as final. Same with the `FileAttachment` on line 376

serious6:
@barisaydinoglu it would be great if you also would add some unittests for the behavior

src/main/java/microsoft/exchange/webservices/data/AttachmentCollection.java
barisaydinoglu: it's wrong. the initial value of contactPhotoFound=false.
the bool set to true, when only one contact photo is found.
if there is multi contact photo an exception is thrown.

barisaydinoglu:
I've fixed if conditions.

src/main/java/microsoft/exchange/webservices/data/AttachmentCollection.java
serious6: ah sry. Didnt see the loop in this case. You are right.

serious6:
@barisaydinoglu are you planning any changes on this PR?

Issue #160, opened 2015-01-06
Requests executing slow due to Kerberos authentication. Mechanism level: Failed to find any Kerberos tgt


rlambrechts:
Apparently when using the httpclient version 4.3.6 I got authentication failures due to negotiating which authentication method to be used. Don't know if it is dependent on  environment or not.

<dependency>
<groupId>org.apache.httpcomponents</groupId>
<artifactId>httpclient</artifactId>
<version>4.3.6</version>
</dependency>

Fixed it by editing the HttpClientWebRequest.java class
public void prepareConnection() throws EWSHttpException {
...
// added:
ArrayList<String> authPrefs = new ArrayList<String>();
      authPrefs.add(AuthSchemes.NTLM);
      rcBuilder.setTargetPreferredAuthSchemes(authPrefs).build();

....
}

serious6:
hi @rlambrechts thanks for the suggestion. Can you create a [PR](https://help.github.com/articles/using-pull-requests/) with the mentioned changes?

Issue #144, opened 2014-12-18
Keep getting an authentication warning


bureaquete:
We are using a Basic authentication, but ews tries using Kerberos and thus gives a warning from HTTPAuthenticator class after each action. It is as follows; 

**NEGOTIATE authentication error: No valid credentials provided (Mechanism level: No valid credentials provided (Mechanism level: Failed to find any Kerberos tgt))**

We initialize ews as follows;

```ExchangeService exchangeWebService = new ExchangeService(ExchangeWebServicesConstants.VERSION);
exchangeWebService.setCredentials(new WebCredentials(username, password, domain));
exchangeWebService.setUrl(ExchangeWebServicesConstants.EXCHANGE_WEB_SERVICE_SERVER_URL);```

How can we get rid of this warning, and stop ews from trying this type of authentication?

rlambrechts:
bureaquete, could you verify if it works with the changes I made in: https://github.com/rlambrechts/ews-java-api

Kind regards,

bureaquete:
@rlambrechts , that did the trick!

Thanks :+1:

PR #164, opened 2015-01-07
NEGOTIATE authentication error: No valid credentials provided (Mechanism level: No valid credentials provided (Mechanism level: Failed to find any Kerberos tgt))

rlambrechts:
Fix issue for #144 and issue #160 (duplicate)

Confirmed to be working: https://github.com/OfficeDev/ews-java-api/issues/144

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/HttpClientWebRequest.java
+++ b/src/main/java/microsoft/exchange/webservices/data/HttpClientWebRequest.java
@@ -10,6 +10,19 @@
 
 package microsoft.exchange.webservices.data;
 
+import java.io.BufferedInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutorService;
+
+import javax.net.ssl.TrustManager;
+
 import org.apache.http.Header;
 import org.apache.http.HttpHost;
 import org.apache.http.HttpResponse;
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@rlambrechts__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

src/main/java/microsoft/exchange/webservices/data/HttpClientWebRequest.java
```diff
@@ -127,6 +138,14 @@ public void prepareConnection() throws EWSHttpException {
...
       rcBuilder.setRedirectsEnabled(isAllowAutoRedirect());
       rcBuilder.setSocketTimeout(getTimeout());
+	  
+	  // fix issue #144 + #160: if we used NTCredentials from above: these are NT credentials
+      if (getUserName() != null) {
+    	ArrayList<String> authPrefs = new ArrayList<String>();
+    	authPrefs.add(AuthSchemes.NTLM);
+    	rcBuilder.setTargetPreferredAuthSchemes(authPrefs).build();
+      }
+      //
```
serious6: * should this behavior also be added to `prepareAsyncConnection`? 
* `rcBuilder.build()` is called twice ref. [l146](https://github.com/rlambrechts/ews-java-api/blob/master/src/main/java/microsoft/exchange/webservices/data/HttpClientWebRequest.java#L146) / [l149](https://github.com/rlambrechts/ews-java-api/blob/master/src/main/java/microsoft/exchange/webservices/data/HttpClientWebRequest.java#L149) is this usefull?
* please align the comment in [l142](https://github.com/rlambrechts/ews-java-api/blob/master/src/main/java/microsoft/exchange/webservices/data/HttpClientWebRequest.java#L142)

rlambrechts: - prepareAsyncConnection: I assule so however I haven't verified this.
- build() => fixed.

src/main/java/microsoft/exchange/webservices/data/HttpClientWebRequest.java
```diff
@@ -189,6 +208,13 @@ public void prepareAsyncConnection() throws EWSHttpException {
       rcBuilder.setConnectionRequestTimeout(getTimeout());
       rcBuilder.setConnectTimeout(getTimeout());
       rcBuilder.setSocketTimeout(getTimeout());
+	  
+      // fix issue #144 + #160: if we used NTCredentials from above: these are NT credentials
+      ArrayList<String> authPrefs = new ArrayList<String>();
```
barisaydinoglu: prepareAsyncConnection doesn't have same controls with prepareConnection.
```if (getUserName() != null) ... ```

serious6:
@rlambrechts thanks for the contribution. If you have adressed the above comment and there are no other suggestions we can have this merged. Maybe you could also confirm the behavior for the async request?

MikeN123:
This PR seems to disable Basic auth, which means this library won't work with O365. I'm currently adding a commit to #180 (which conflicts with this PR anyway), to enable both NTLM and Basic auth, on target and proxy.

vboctor:
Thanks @rlambrechts for your contribution.  I've just merged #180 which included the port of this fix by @MikeN123.  Enabling basic auth will keep the Office 365 access working.

PR #165, opened 2015-01-07
Switched to more robust parsing of start-date of recurring events

ahaarrestad:
Removed simple date-parser and switched start-date parsing for recurring events to the more robust parser.

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/ExchangeServiceBase.java
+++ b/src/main/java/microsoft/exchange/webservices/data/ExchangeServiceBase.java
@@ -38,8 +38,6 @@ import java.util.regex.Pattern;
  */
 public abstract class ExchangeServiceBase {
 
-  // Fields
-
   /**
    * Prefix for "extended" headers.
    */
@@ -98,8 +96,7 @@ public abstract class ExchangeServiceBase {
   /**
    * The requested server version.
    */
-  private ExchangeVersion requestedServerVersion =
-      ExchangeVersion.Exchange2010_SP2;
+  private ExchangeVersion requestedServerVersion = ExchangeVersion.Exchange2010_SP2;
 
   /**
    * The server info.
@@ -109,8 +106,6 @@ public abstract class ExchangeServiceBase {
   private Map<String, String> httpHeaders = new HashMap<String, String>();
 
   private Map<String, String> httpResponseHeaders = new HashMap<String, String>();
-
-  private TimeZone timeZone;
 
   private WebProxy webProxy;
 
@@ -123,34 +118,21 @@ public abstract class ExchangeServiceBase {
   // protected static HttpStatusCode AccountIsLocked = (HttpStatusCode)456;
 
   /**
-   * Static members
-   */
-
-  protected HttpClientConnectionManager getSimpleHttpConnectionManager() {
-    return httpConnectionManager;
-  }
-
-  /**
    * Default UserAgent.
    */
-  private static String defaultUserAgent = "ExchangeServicesClient/" +
-      EwsUtilities.getBuildVersion();
+  private static String defaultUserAgent = "ExchangeServicesClient/" + EwsUtilities.getBuildVersion();
 
-  /**
-   * @return TimeZone
-   */
-  private TimeZone getTimeZone() {
-    return this.timeZone;
+  protected HttpClientConnectionManager getHttpConnectionManager() {
+    return httpConnectionManager;
   }
 
   /**
    * Initializes a new instance.
    *
-   * @param requestedServerVersion The requested server version.
+   * This constructor performs the initialization of the HTTP connection manager, so it should be called by
+   * every other constructor.
    */
   protected ExchangeServiceBase() {
-    //real init -- ONLY WAY TO BUILD AN OBJECT => each constructor must call this() to build properly the httpConnectionManager
-    this.timeZone = TimeZone.getDefault();
     this.setUseDefaultCredentials(true);
 
     try {
@@ -166,15 +148,8 @@ public abstract class ExchangeServiceBase {
   }
 
   protected ExchangeServiceBase(ExchangeVersion requestedServerVersion) {
-    // Removed because TimeZone class in Java doesn't maintaining the
-    //history of time change rules for a given time zone
-    //this(requestedServerVersion, TimeZone.getDefault());
     this();
     this.requestedServerVersion = requestedServerVersion;
-  }
-
-  protected ExchangeServiceBase(ExchangeServiceBase service) {
-    this(service, service.getRequestedServerVersion());
   }
 
   protected ExchangeServiceBase(ExchangeServiceBase service, ExchangeVersion requestedServerVersion) {
@@ -188,14 +163,7 @@ public abstract class ExchangeServiceBase {
     this.preAuthenticate = service.isPreAuthenticate();
     this.userAgent = service.getUserAgent();
     this.acceptGzipEncoding = service.getAcceptGzipEncoding();
-    this.timeZone = service.getTimeZone();
     this.httpHeaders = service.getHttpHeaders();
-  }
-
-
-  protected ExchangeServiceBase(ExchangeVersion requestedServerVersion, TimeZone timezone) {
-    this(requestedServerVersion);
-    this.timeZone = timezone;
   }
 
   // Event handlers
@@ -206,9 +174,9 @@ public abstract class ExchangeServiceBase {
    * @param writer The XmlWriter to which to write the custom SOAP headers.
    */
   protected void doOnSerializeCustomSoapHeaders(XMLStreamWriter writer) {
-    EwsUtilities.EwsAssert(writer != null,
-        "ExchangeService.DoOnSerializeCustomSoapHeaders",
-        "writer is null");
+    EwsUtilities
+        .EwsAssert(writer != null, "ExchangeService.DoOnSerializeCustomSoapHeaders", "writer is null");
+
     if (null != getOnSerializeCustomSoapHeaders() &&
         !getOnSerializeCustomSoapHeaders().isEmpty()) {
       for (ICustomXmlSerialization customSerialization : getOnSerializeCustomSoapHeaders()) {
@@ -231,9 +199,8 @@ public abstract class ExchangeServiceBase {
    * @throws ServiceLocalException       the service local exception
    * @throws java.net.URISyntaxException the uRI syntax exception
    */
-  protected HttpWebRequest prepareHttpWebRequestForUrl(URI url,
-      boolean acceptGzipEncoding, boolean allowAutoRedirect)
-      throws ServiceLocalException, URISyntaxException {
+  protected HttpWebRequest prepareHttpWebRequestForUrl(URI url, boolean acceptGzipEncoding,
+    boolean allowAutoRedirect) throws ServiceLocalException, URISyntaxException {
     // Verify that the protocol is something that we can handle
     if (!url.getScheme().equalsIgnoreCase("HTTP")
         && !url.getScheme().equalsIgnoreCase("HTTPS")) {
@@ -306,24 +273,16 @@ public abstract class ExchangeServiceBase {
    *
    * @throws Exception
    */
-  protected void internalProcessHttpErrorResponse(
-      HttpWebRequest httpWebResponse,
-      Exception webException,
-      TraceFlags responseHeadersTraceFlag,
-      TraceFlags responseTraceFlag) throws Exception {
-    EwsUtilities.EwsAssert(
-        500 != httpWebResponse.getResponseCode(),
+  protected void internalProcessHttpErrorResponse(HttpWebRequest httpWebResponse, Exception webException,
+      TraceFlags responseHeadersTraceFlag, TraceFlags responseTraceFlag) throws Exception {
+    EwsUtilities.EwsAssert(500 != httpWebResponse.getResponseCode(),
         "ExchangeServiceBase.InternalProcessHttpErrorResponse",
-        "InternalProcessHttpErrorResponse does not handle 500 ISE errors," +
-            " the caller is supposed to handle this.");
+        "InternalProcessHttpErrorResponse does not handle 500 ISE errors, the caller is supposed to handle this.");
 
-    this.processHttpResponseHeaders(responseHeadersTraceFlag,
-        httpWebResponse);
+    this.processHttpResponseHeaders(responseHeadersTraceFlag, httpWebResponse);
 
-    // E14:321785 -- Deal with new HTTP
-    // error code indicating that account is locked.
-    // The "unlock" URL is returned as
-    // the status description in the response.
+    // E14:321785 -- Deal with new HTTP error code indicating that account is locked.
+    // The "unlock" URL is returned as the status description in the response.
     if (httpWebResponse.getResponseCode() == 456) {
       String location = httpWebResponse.getResponseContentType();
 
@@ -332,13 +291,11 @@ public abstract class ExchangeServiceBase {
         accountUnlockUrl = new URI(location);
       }
 
-      this.traceMessage(responseTraceFlag, String.format("Account is locked." +
-          " Unlock URL is {0}", accountUnlockUrl));
+      this.traceMessage(responseTraceFlag,
+          String.format("Account is locked. Unlock URL is {0}", accountUnlockUrl));
 
-      throw new AccountIsLockedException(
-          String.format(Strings.AccountIsLocked, accountUnlockUrl),
-          accountUnlockUrl,
-          webException);
+      throw new AccountIsLockedException(String.format(Strings.AccountIsLocked, accountUnlockUrl),
+          accountUnlockUrl, webException);
     }
   }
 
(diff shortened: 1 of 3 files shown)


azurecla:
Hi __@ahaarrestad__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

azurecla:
__@ahaarrestad__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

serious6:
hi @ahaarrestad can you state why this change is important and have an issue created? I think there should also be a unittest added to clarify the expected parsing result.

ahaarrestad:
I've added issue https://github.com/OfficeDev/ews-java-api/issues/169 to explain the problem. I'll look into a couple of unit-tests to make sure that nothing has been broken by moving to the more robust date-parser.

ahaarrestad:
I will rewrite the pull-request a bit to make sure we don't break anything. The parsing ended up an hour or two off. I'll add the appropriate unit-tests and a fix to make it all ok again.

ahaarrestad:
Rewrote the pull-request and added a simple test for parsing dates

src/test/java/microsoft/exchange/webservices/data/RecurringAppointmentDateParsing.java
```diff
@@ -0,0 +1,42 @@
...
+		Date newParsing = exchangeServiceBaseMock.convertStartDateToUnspecifiedDateTime(dateString);
+		Date oldParsing = convertStartDateToUnspecifiedDateTime(dateString);
+		assertEquals(oldParsing, newParsing);
+	}
+
+	@Test
+	public void parsePreviouslyUnbarsableRecurrenceDateFormat() {
+		String dateString = "2015-01-08-01:00";
+		Date parsed = exchangeServiceBaseMock.convertUniversalDateTimeStringToDate(dateString);
+		assertNotNull(parsed);
```
serious6: maybe we should also assert the correct parsing here? Something like:
* year is 2015
* month / day are valid after parsing
* hour / minutes are correct too

ahaarrestad:
I have extended the unit-tests to verify parsing of a couple of simple examples.

src/test/java/microsoft/exchange/webservices/data/RecurringAppointmentDateParsing.java
```diff
@@ -0,0 +1,81 @@
+package microsoft.exchange.webservices.data;
```
serious6: License header is missing

src/test/java/microsoft/exchange/webservices/data/RecurringAppointmentDateParsing.java
```diff
@@ -0,0 +1,81 @@
...
+
+import java.text.DateFormat;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Calendar;
+import java.util.Date;
+import java.util.GregorianCalendar;
+
+import org.junit.Test;
+
```
serious6: Missing annotation @runwith

src/test/java/microsoft/exchange/webservices/data/RecurringAppointmentDateParsing.java
```diff
@@ -0,0 +1,81 @@
...
+public class RecurringAppointmentDateParsing extends BaseTest {
```
serious6: File doesn't match name pattern: <name>Test.java

src/test/java/microsoft/exchange/webservices/data/RecurringAppointmentDateParsing.java
```diff
@@ -0,0 +1,81 @@
...
+
+
+	@Test
```
serious6: Please add javaDoc to methods and the class

serious6:
I have added some formal suggestions for handling the test. As I saw you removed the old parsing method. Maybe it could be useful to just declare it as deprecated?

ahaarrestad:
I'll update the headers and javadoc. We do not need to deprecate the methode, since it was not used outside the api (protected) and I have changed the only method calling it from within the api. Agree?

src/main/java/microsoft/exchange/webservices/data/ExchangeServiceBase.java
```diff
@@ -471,28 +479,35 @@ protected Date convertUniversalDateTimeStringToDate(String dateString) {
     if (dateString == null || dateString.isEmpty()) {
       return null;
     } else {
-      if (dateString.endsWith("Z")) {
+    TimeZone UTCTimeZone = TimeZone.getTimeZone("UTC");
```
serious6: This is really hard to understand. If I understand it right the only difference in the parsing is the changed pattern when a `ParseException` occurs. Maybe we should have this refactored. I would suggest extracting the parsing process into another method. Smth like:
```java
private String doParsing(final boolean useUtcParser, final Date dateToBeParsed, final String pattern) 
   throws ParseException {
  if(useUtcParser) {
      	utcFormatter.setTimeZone(TimeZone.getTimeZone("UTC"));
  }
  return new SimpleDateFormat(utcPattern).parse(dateToBeParsed);
}
```
I also would suggest to let this happen in a loop rather then multiple try/catch. This loop may run until all patterns have been used or the parsing ends correctly.

serious6: I also would suggest in adding more comments to the things being done. This can be usefull if someone else will look at the code later on.

src/main/java/microsoft/exchange/webservices/data/ExchangeServiceBase.java
```diff
@@ -456,7 +456,15 @@ private void traceHttpResponseHeaders(TraceFlags traceType,
...
   protected Date convertUniversalDateTimeStringToDate(String dateString) {
-    String localTimeRegex = "^(.*)([+-]{1}\\d\\d:\\d\\d)$";
+	  return convertDateStringToDate(dateString, true);
+  }
+
+  protected Date convertStartDateToUnspecifiedDateTime(String value) {
+	  return convertDateStringToDate(value, false);
+  }
+	  
+  private Date convertDateStringToDate(String dateString, boolean isUTC) {
```
serious6: javaDoc is missing

src/main/java/microsoft/exchange/webservices/data/ExchangeServiceBase.java
```diff
@@ -456,7 +456,15 @@ private void traceHttpResponseHeaders(TraceFlags traceType,
    * @return DateTime Returned date is always in UTC date.
    */
   protected Date convertUniversalDateTimeStringToDate(String dateString) {
-    String localTimeRegex = "^(.*)([+-]{1}\\d\\d:\\d\\d)$";
+	  return convertDateStringToDate(dateString, true);
+  }
+
+  protected Date convertStartDateToUnspecifiedDateTime(String value) {
+	  return convertDateStringToDate(value, false);
+  }
```
serious6: I would suggest the two above methods being removed. Since the developer can be forced to decide if it is an UTC String or not (in my opinion). `convertDateStringToDate` may be renamed to `convertUniversalDateTimeStringToDate` again I think.

src/test/java/microsoft/exchange/webservices/data/RecurringAppointmentDateParsingTest.java
```diff
@@ -0,0 +1,114 @@
...
+		Calendar calendar = new GregorianCalendar(TimeZone.getTimeZone("Europe/Paris"));
+		calendar.setTime(parsed);
+		assertEquals(2015, calendar.get(Calendar.YEAR));
+		assertEquals(0, calendar.get(Calendar.MONTH));
+		assertEquals(8, calendar.get(Calendar.DATE));
+		assertEquals(2, calendar.get(Calendar.HOUR));
+		assertEquals(0, calendar.get(Calendar.MINUTE));
+
+	}
+
```
serious6: please add more tests for a better code coverage (I think every pattern should be tested)

src/test/java/microsoft/exchange/webservices/data/RecurringAppointmentDateParsingTest.java
```diff
@@ -0,0 +1,114 @@
...
+		String dateString = "2015-01-08-01:00";
+		Date parsed = exchangeServiceBaseMock.convertUniversalDateTimeStringToDate(dateString);
+		assertNotNull(parsed);
+		Calendar calendar = new GregorianCalendar(TimeZone.getTimeZone("Europe/Paris"));
+		calendar.setTime(parsed);
+		assertEquals(2015, calendar.get(Calendar.YEAR));
+		assertEquals(0, calendar.get(Calendar.MONTH));
+		assertEquals(8, calendar.get(Calendar.DATE));
+		assertEquals(2, calendar.get(Calendar.HOUR));
+		assertEquals(0, calendar.get(Calendar.MINUTE));
```
serious6: maybe I should have looked at this in more detail but are the expected results really correct (same for `parsePreviouslyUnbarsableRecurrenceDateFormat2`)?

ahaarrestad: The code works fine. I have tested it on real data as well, and the parsed results match the outlook calendar for all the different timezone-formats used in our code.

ahaarrestad:
I have updated the pull-request as we have discussed. Please review the changes. Anything more we should change before merging?

src/test/java/microsoft/exchange/webservices/data/RecurringAppointmentDateParsingTest.java
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  @Test
+  public void parseLocalPattern2NoDefault() {
+    String dateString = "2015-01-08T10:11:12"; // "yyyy-MM-dd'T'HH:mm:ss"
+    Date parsed = exchangeServiceBaseMock.convertDateTimeStringToDate(dateString, false);
+    Calendar calendar = new GregorianCalendar();
+    calendar.setTime(parsed);
+    assertEquals(2015, calendar.get(Calendar.YEAR));
+    assertEquals(0, calendar.get(Calendar.MONTH));
```
serious6: why is it `0` here? since the String was _2015-**01**-08_.

src/test/java/microsoft/exchange/webservices/data/RecurringAppointmentDateParsingTest.java
```diff
@@ -0,0 +1,139 @@
...
+   */
+  @Test
+  public void parseAPIRecurrenceDateFormat() throws ParseException {
+    String dateString = "2015-01-08Z";
+    Date newParsing = exchangeServiceBaseMock.convertDateTimeStringToDate(dateString, false);
+    Date oldParsing = convertStartDateToUnspecifiedDateTime(dateString);
+    assertEquals(oldParsing, newParsing);
+  }
+
+  @Test
```
serious6: new tests are missing javadoc :-)

src/main/java/microsoft/exchange/webservices/data/ExchangeServiceBase.java
```diff
@@ -554,21 +507,30 @@ protected Date convertUniversalDateTimeStringToDate(String dateString) {
...
    */
-  protected Date convertStartDateToUnspecifiedDateTime(String value)
-      throws ParseException {
-    if (value == null || value.isEmpty()) {
-      return null;
-    } else {
-      DateFormat df = new SimpleDateFormat("yyyy-MM-dd'Z'");
-      return df.parse(value);
+  Date parseDateTimeString(String dateString, boolean isUTC, String... patterns) {
+    ParseException lastException = null;
```
serious6: hmm do you think this is really tracable in the event of an error? iterating through the patterns will only inform the user if all patterns result in a `ParseException`. Let´s assume we have 3 patterns. 
* pattern 1 should fail with the input
* pattern 2 should match
* pattern 3 should fail with the input

now for some reasons pattern 2 also fails. Now the user will only get noticed that the input fails parsing with pattern 3.

serious6: maybe someone else has suggestions here too?

src/test/java/microsoft/exchange/webservices/data/RecurringAppointmentDateParsingTest.java
ahaarrestad: http://stackoverflow.com/questions/344380/why-is-january-month-0-in-java-calendar

src/main/java/microsoft/exchange/webservices/data/ExchangeServiceBase.java
ahaarrestad: This is the same error the user would get today, thus not introducing "new" errors. But I agree that it's not an optimal solution.

I will look into an aggregated result.

src/test/java/microsoft/exchange/webservices/data/RecurringAppointmentDateParsingTest.java
ahaarrestad: fixed javadoc

src/main/java/microsoft/exchange/webservices/data/ExchangeServiceBase.java
ahaarrestad: I have now added aggregated results to explain more which patterns were tried when parsing.

ahaarrestad:
any more changes we need to address?

src/main/java/microsoft/exchange/webservices/data/ExchangeServiceBase.java
```diff
@@ -554,21 +507,29 @@ protected Date convertUniversalDateTimeStringToDate(String dateString) {
...
+  Date parseDateTimeString(String dateString, boolean isUTC, String... patterns) {
+    for (String pattern : patterns) {
+      SimpleDateFormat utcFormatter = new SimpleDateFormat(pattern);
+      if (isUTC) {
+        utcFormatter.setTimeZone(TimeZone.getTimeZone("UTC"));
+      }
+      try {
+        return utcFormatter.parse(dateString);
+      } catch (ParseException e) {
+        //do nothing.
```
serious6: maybe we can add some logging here?

ahaarrestad: I will look into logging here. Also remove the stdOut message?

ahaarrestad:
fixed merge-conflict

ahaarrestad:
anything more we should do regarding this pull request?

serious6:
@ahaarrestad thanks for your work. 
If no one else has any comments to adress we are fine with this I think. Maybe you can squash the unnecessary commits only leaving the ones relevant for understanding the flow (Codechanges seperated from Tests for instance)

evpaassen:
I'd like to review this one too. This date parse routine is quite a crucial part of the library which has contained several bugs before. I'll review these changes ASAP.

ahaarrestad:
I'm working on a rebase / squash. something went wrong.... I'll be right back :-)

ahaarrestad:
can anybody please reopen this pull-request?

evpaassen:
@ahaarrestad Your branch currently doesn't contain any new commits over the OfficeDev/ews-java-api master branch.

evpaassen:
If you need any help, I'm on Gitter.

ahaarrestad:
Do you want me to open a new pull-request or can this one be re-opened?

serious6:
You need to commit on the branch this PR is connected with. I'll reopen it then.

ahaarrestad:
Commits made earlier today.

evpaassen:
An alternative approach: https://github.com/eveoh/ews-java-api/compare/fix-date-parsing?expand=1

serious6:
I was wondering that this PR is against your master branch maybe thats the cause of the problem while it cannot be reopened? [seen here](http://jasonnoble.org/2012/09/github-unable-to-reopen-a-closed-pull-request.html). Maybe just create a new PR. @evpaassen the alternative approach has no tests included.

evpaassen:
@serious6 http://stackoverflow.com/questions/5601730/should-private-protected-methods-be-under-unit-test

ahaarrestad:
It was a mistake to use my head-branch. Won't do that next time :-)

ahaarrestad:
regarding the tests: testing protected methods through public methods is ok, but might be tricky. Sometimes you can verify a method better if you test it separately.

serious6:
@evpaassen I think the valid argument for providing tests should be to test a (perhabs important) behavior rather then an implementation detail. The test should underline the expected behavior of the methods and will help us to avoid exceptions in the future. One of the argument in the post you mentioned is that these methods would be covered by tests at higher level. This may be true for projects that already have large test-suites which we really dont have by now. @ahaarrestad I leave it up to you if you want to provide tests for the methods changed or if you want to write tests for the calling methods :-)

evpaassen:
@serious6 The real problem is that the method shouldn't be protected. I'm creating tests for this method a.t.m.

Issue #158, opened 2015-01-02
Errors in project formatter [codeStyleSettings.xml]


serious6:
While working with the formatter we noticed some kind of errors and differences coming around with the styleguide. The following is taken from @barisaydinoglu:

* Comment line formatting:
  * IntelliJ: does not format comment line indentation. So, every indentation of the code above is right based on your formatter.
  * Eclipse: formats (or changes) comment line indentation.
* google-styleguide for Java works different for Eclipse and Intellij.
* Mix of space and tab are being used as comment indentation character.
  * A tab could be a different number of columns depending on your environment, but a space is always one column.
  * In terms of how many spaces (or tabs) constitutes indentation, it's more important to be consistent throughout your code than to use any specific tab stop value.

serious6:
@barisaydinoglu thanks for your support.

PR #166, opened 2015-01-07
ExchangeServiceBase cleanup

evpaassen:
This cleans up the ExchangeServiceBase class before I attempt to fix the shutdown of the HTTP connection manager. Most of it concerns formatting, as the previous formatting seems to be performed with 'keep line breaks' on.

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/ExchangeServiceBase.java
+++ b/src/main/java/microsoft/exchange/webservices/data/ExchangeServiceBase.java
@@ -110,8 +110,6 @@ public abstract class ExchangeServiceBase {
 
   private Map<String, String> httpResponseHeaders = new HashMap<String, String>();
 
-  private TimeZone timeZone;
-
   private WebProxy webProxy;
 
   private HttpClientConnectionManager httpConnectionManager;
@@ -137,20 +135,12 @@ public abstract class ExchangeServiceBase {
       EwsUtilities.getBuildVersion();
 
   /**
-   * @return TimeZone
-   */
-  private TimeZone getTimeZone() {
-    return this.timeZone;
-  }
-
-  /**
    * Initializes a new instance.
    *
    * @param requestedServerVersion The requested server version.
    */
   protected ExchangeServiceBase() {
     //real init -- ONLY WAY TO BUILD AN OBJECT => each constructor must call this() to build properly the httpConnectionManager
-    this.timeZone = TimeZone.getDefault();
     this.setUseDefaultCredentials(true);
 
     try {
@@ -166,9 +156,6 @@ public abstract class ExchangeServiceBase {
   }
 
   protected ExchangeServiceBase(ExchangeVersion requestedServerVersion) {
-    // Removed because TimeZone class in Java doesn't maintaining the
-    //history of time change rules for a given time zone
-    //this(requestedServerVersion, TimeZone.getDefault());
     this();
     this.requestedServerVersion = requestedServerVersion;
   }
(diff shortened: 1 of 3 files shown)


azurecla:
    <span>
    </span>

TTYL, AZPRBOT;

azurecla:
__@evpaassen__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

src/main/java/microsoft/exchange/webservices/data/ExchangeServiceBase.java
```diff
@@ -98,19 +96,16 @@
...
    */
   private ExchangeServerInfo serverInfo;
 
-  private Map<String, String> httpHeaders = new HashMap<String, String>();
-
-  private Map<String, String> httpResponseHeaders = new HashMap<String, String>();
+  private Map<String, String> httpHeaders = new HashMap<>();
 
-  private TimeZone timeZone;
+  private Map<String, String> httpResponseHeaders = new HashMap<>();
```
serious6: empty diamond operator causes the build to fail because it is a 1.7 feature. Maybe we should avoid this in the use of compatibility since it is just syntactic sugar.

evpaassen: I reverted this specific change, but the project settings indicate I can use Java 1.7 features, but Maven is not set up correctly. I have created a separate issue for this (#167).

src/main/java/microsoft/exchange/webservices/data/ExchangeServiceBase.java
```diff
@@ -593,81 +540,69 @@ protected void setCustomUserAgent(String userAgent) {
...
+//        String v = iter.next();
+//
+//        if (cookieValue == null)
+//          cookieValue = v;
+//        else
+//          cookieValue = cookieValue + ";" + v;
+//      }
+//    }
+//    return cookieValue;
+//  }
```
serious6: shouldn't this be removed in the order of cleaning up?

serious6:
ref. #158 cause of the formatting.

src/main/java/microsoft/exchange/webservices/data/ExchangeServiceBase.java
evpaassen: I don't really know what's the deal with this code. There's a lot out code that has been commented out in this library, most of which could probably be removed...

serious6:
@evpaassen are you planning to add more changes to this PR? removal of commented code for instance? I think this could be merged if the answer is no.

evpaassen:
I'm currently working on some issues with this class (how it uses the HTTP connection manager and client), but I'll submit another PR for that, as that one involves rewriting rather than trivial clean up. That one also involves cookie handling, so I'll have a look at the commented methods then.

I have rebased the branch on the current `master`, so this one can be merged now.

serious6:
_merged with rev. 99bec6039f5ade4824c7132c872752c7b391f201_
**Thanks for contribution**

evpaassen:
@serious6, thanks for merging this, I really appreciate that. But I see all commits have been squashed into a single commit. I think this is not really useful as those were logical commits, refactoring the class step by step. It's much harder browse the history of the file now.

I rebased the branch, so it was mergeable with the merge button. Also, when a PR is merged with the merge button, the pull request gets the 'merged' state instead of 'closed', which is really useful when browsing  the list of pull requests. Is there a reason why PRs in this repo are not merged by using this button?

serious6:
I squashed it in the use of having a better everview in commits. Maybe we can differ from this behavior in the future if commits are important for the history. This could also be done if we force users to squash unecessary commits. Think this should be added to **Contributing.md** same as the need of adding tests for the PR.

evpaassen:
I agree unnecessary commits should be squashed, but commits should stay logical and preferably small, so they're easy to understand.

Issue #167, opened 2015-01-07
Java language level in Maven pom file and IntelliJ project config don't match


evpaassen:
The Java language level in the IntelliJ IDEA config `.idea/misc.xml` is set to `1.7.`, while Maven assumes `1.5`.

The following line is in the `pom.xml`, but it doesn't seem to do anything and I don't know where it originates from. It was in the initial commit.
```
<javaLanguage.version>1.7</javaLanguage.version>
```

serious6:
hi @evpaassen its because the missing maven-compiler-plugin definition in `pom.xml`
```xml
[...]
    <build>
        <plugins>
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-compiler-plugin</artifactId>
                <configuration>
                    <encoding>${project.build.sourceEncoding}</encoding>
                    <source>${javaLanguage.version}</source>
                    <target>${javaLanguage.version}</target>
                </configuration>
            </plugin>
[...]
```
if this would be added the compiler would be set to the specified level.
I think for detecting downgrade issues in the compiler we also should add the following to the `.travis.yml`
```xml
language: java
jdk:
  - oraclejdk8
  - oraclejdk7
  - oraclejdk6
[...]
```
this will help us to keep track of older and newer builds.

barisaydinoglu:
jdk 1.7 is not necessary, 1.6+ is enough for ews-java-api.

i'm currently using with jdk1.6.
i cant see a reason to update bytecode version upgrade.

pathikrit:
I don't think we should check-in editor configs in this repo. We should add this to gitignore: https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore

evpaassen:
@pathikrit Please see https://www.jetbrains.com/idea/help/managing-projects-under-version-control.html. The .gitignore file you mention is correct and it does **not** exclude all IntelliJ editor configuration.

PR #168, opened 2015-01-07
Fix Java language level in pom.xml

evpaassen:
IntelliJ IDEA project files have also been updated by performing a Maven refresh.

Fixes OfficeDev/ews-java-api#167.

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -10,8 +10,7 @@
       </list>
     </option>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_7" assert-keyword="true" jdk-15="true" project-jdk-name="1.7" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_6" assert-keyword="true" jdk-15="true" project-jdk-name="1.7" project-jdk-type="JavaSDK">
     <output url="file://$PROJECT_DIR$/out" />
   </component>
 </project>
-
--- a/ews-java-api.iml
+++ b/ews-java-api.iml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <module org.jetbrains.idea.maven.project.MavenProjectsManager.isMavenModule="true" type="JAVA_MODULE" version="4">
-  <component name="NewModuleRootManager" inherit-compiler-output="false">
+  <component name="NewModuleRootManager" LANGUAGE_LEVEL="JDK_1_6" inherit-compiler-output="false">
     <output url="file://$MODULE_DIR$/target/classes" />
     <output-test url="file://$MODULE_DIR$/target/test-classes" />
     <content url="file://$MODULE_DIR$">
@@ -11,13 +11,12 @@
     <orderEntry type="inheritedJdk" />
     <orderEntry type="sourceFolder" forTests="false" />
     <orderEntry type="library" name="Maven: org.apache.httpcomponents:httpclient:4.3.5" level="project" />
+    <orderEntry type="library" name="Maven: commons-codec:commons-codec:1.6" level="project" />
     <orderEntry type="library" name="Maven: org.apache.httpcomponents:httpcore:4.3.3" level="project" />
     <orderEntry type="library" name="Maven: commons-logging:commons-logging:1.2" level="project" />
-    <orderEntry type="library" name="Maven: commons-codec:commons-codec:1.6" level="project" />
     <orderEntry type="library" name="Maven: jcifs:jcifs:1.3.17" level="project" />
     <orderEntry type="library" scope="TEST" name="Maven: junit:junit:4.11" level="project" />
     <orderEntry type="library" scope="TEST" name="Maven: org.hamcrest:hamcrest-core:1.3" level="project" />
     <orderEntry type="library" scope="TEST" name="Maven: org.hamcrest:hamcrest-all:1.3" level="project" />
   </component>
 </module>
-
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
         <!-- Eliminates the file encoding warning. Of course, all of your files
             should probably be UTF-8 nowadays. -->
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <javaLanguage.version>1.7</javaLanguage.version>
+        <javaLanguage.version>1.6</javaLanguage.version>
     </properties>
 
     <dependencies>
@@ -84,6 +84,15 @@
 
     <build>
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <encoding>${project.build.sourceEncoding}</encoding>
+                    <source>${javaLanguage.version}</source>
+                    <target>${javaLanguage.version}</target>
+                </configuration>
+            </plugin>
             <plugin>
                 <groupId>org.apache.felix</groupId>
                 <artifactId>maven-bundle-plugin</artifactId>
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -25,6 +25,8 @@
         <module name="ews-java-api" />
       </profile>
     </annotationProcessing>
+    <bytecodeTargetLevel>
+      <module name="ews-java-api" target="1.6" />
+    </bytecodeTargetLevel>
   </component>
 </project>
-


azurecla:

TTYL, AZPRBOT;

ews-java-api.iml
```diff
@@ -1,6 +1,18 @@
...
+      <configuration manifestGenerationMode="OsmorcControlled" manifestLocation="" jarfileLocation="ews-java-api-1.3-SNAPSHOT.jar" outputPathType="CompilerOutputPath" bndFileLocation="" bundlorFileLocation="" bundleActivator="" bundleSymbolicName="com.microsoft.office.ews-java-api" bundleVersion="1.3.0.SNAPSHOT" ignoreFilePattern="" useProjectDefaultManifestFileLocation="true" alwaysRebuildBundleJAR="false" doNotSynchronizeWithMaven="false">
+        <additionalProperties>
+          <property key="Bundle-Description" value="Exchange Web Services (EWS) Java API" />
+          <property key="Bundle-Vendor" value="Microsoft" />
+          <property key="Bundle-DocURL" value="http://www.microsoft.com/" />
+        </additionalProperties>
+        <additionalJARContents />
+      </configuration>
+    </facet>
+  </component>
```
serious6: the change from line 3 - 14 ("FacetManager") may possibly not refer to the compiler-level?

pom.xml
```diff
@@ -85,6 +85,15 @@
     <build>
         <plugins>
             <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.2</version>
+                <configuration>
+                    <source>1.7</source>
+                    <target>1.7</target>
```
serious6: 92 - 93 perhabs we should use the `${javaLanguage.version}` here.

.idea/misc.xml
```diff
@@ -3,6 +3,9 @@
   <component name="EntryPointsManager">
     <entry_points version="2.0" />
   </component>
+  <component name="FrameworkDetectionExcludesConfiguration">
+    <file type="Osmorc" url="file://$PROJECT_DIR$" />
+  </component>
```
serious6: Is this also about compiler-level?

ews-java-api.iml
evpaassen: No it doesn't, but this file is being generated by IntelliJ IDEA, and this change originates from #98, which lacks the update of this file.

pom.xml
serious6: I was wondering if there is the posibility to inherit the used jdk from the environment. This would allow us to check the build against other jdk´s as well. For example in travis:
`.travis.yml`
```yml
language: java
jdk:
  - oraclejdk8
  - oraclejdk7
  - oraclejdk6
[...] 
```

evpaassen: To get a reproducable build, we should really set the source and target version. That should not prevent Travis CI compiling it with JDK8, I think. Compiling it with JDK6 won't work anyway, if the language level is set to 1.7, which allows us to use Java 7 features.

evpaassen:
I updated the PR to address your comment about the `javaLanguage.version` property.

ews-java-api.iml
serious6: ok. But as I see there is some information added like the jar-filename. This will result in once the version-number will be changed this file needs to be changed too. IntelliJ will also just update the file if the project is launched with Osmorc plugin enabled etc.

barisaydinoglu:
jdk 1.7 is not necessary, 1.6 is enough for ews-java-api.

evpaassen:
The language level of the project is not for me to decide. I only corrected the pom.xml to be consistent. Perhaps @vboctor could have a look at this. The current codebase compiles just fine against a language level of 1.6, but we wouldn't be able to use Java 7 features like diamond operators, exception multi-catch, switch on strings.

ews-java-api.iml
evpaassen: Then it will be best to update this file after every version change, I guess. Another option would be to put this file in `.gitignore` and add this file with force when it needs updating. Otherwise every contributor will need to exclude this file from new commits, as IntelliJ will keep adding these lines.

The option of adding this file to `.gitignore` is a dangerous one, I think, as it won't be clear when the file does need updating in the repository.

barisaydinoglu:
Have you searched the source code?
There is no diamond operators, exception multi-catch or switch on string used.

evpaassen:
> Have you searched the source code? There is no diamond operators, exception multi-catch or switch on string used.

Of course, because otherwise it wouldn't compile properly against a language level of 1.6. I only stated we wouldn't be able to use these features in the future. Again, it's not for me to decide. If the project maintainers decide the language level will be 1.6, I'll update the PR accordingly.

But I must say Java 1.6 is EOL for almost two years now.

serious6:
just as some information being added to this topic other apis also support java 1.5+, (JavaBridgetoExchange or JWebServicesForExchange for instance). As you said there should be a project manager´s decision on this.

evpaassen:
I removed the configuration of the OSGi plugin for now, as the plugin seems to display errors in IntelliJ. We'll still need a decision on the language level.

vboctor:
I generally would support the oldest version possible until I have a reason not to.  I don't know if we have a reason where such support is holding us back.  Also I would drop such support ideally after we have a release out, providing a maintenance branch for such Java version.  @serious6 provides a good data point for similar libraries supporting all the way back to 1.5.

PaulThorne:
I'm working with this library through IBM Domino servers, which make use of an IBM modified JVM which is only 1.6 (even in the latest versions of Domino). So if support goes beyond 1.6 it will make life difficult.

ahaarrestad:
Many are stuck with java 1.6 even though it is EOL, thus updgrading libs to higher versjons should be avoided as long as possible.

evpaassen:
OK, so now it clear Java 1.6 support can't be dropped. @PaulThorne supplied a valid reason and there will probably some other platforms too.

But would we also need to support Java 1.5 (EOL since 2009), is anyone still using it? I found that using the 1.5 language level with JDK8 is already deprecated and will be removed entirely with JDK9 (see http://openjdk.java.net/jeps/182).

evpaassen:
I have updated the PR to support Java 1.6. If we also need to support 1.5, please let me know.

serious6:
There is not much difference in 1.5 - 1.6 most is just about incrementing and the definition of `@Override` and some other stuff we will not use in this api anyway. But I think 1.6 is enough of downgrade compatibility :-)

.idea/misc.xml
```diff
@@ -10,8 +10,7 @@
       </list>
     </option>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_7" assert-keyword="true" jdk-15="true" project-jdk-name="1.7" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_6" assert-keyword="true" jdk-15="true" project-jdk-name="1.7" project-jdk-type="JavaSDK">
```
serious6: `project-jdk-name="1.7"` should also be changed to `1.6`

evpaassen: Not necessarily. This is the name of the jdk used to compile the project. It refers to the name someone gives to a configured JDK in the *global* IDE settings and will probably differ per user. Furthermore, someone could perfectly use a 1.7 JDK to compile the project, or even a 1.8 JDK, while targeting 1.6 language level and bytecode version.

Here's what happens: if the user has an JDK configured named `1.7`, that one will be used. If not, IntelliJ will give a clear error, suggesting to pick one of the configured JDKs instead. I think that's fine, won't get any better than this.

B.T.W. `1.7` is the default name IntelliJ gives to a 1.7 JDK.

serious6: sounds reasonable thanks for the post.

serious6:
if there are no other comments on this I think we can have this merged.

serious6:
**Thanks for the contribution**
_merged with cdc30b66f1527d419f33f83fcc41eaa5a95fc83c_

pathikrit:
Probably too late to jump in on this but most people don't usually like editor specific configs (be it IntelliJ or emacs or Eclipse) checked into the repo. There are some standard gitignores here which disallow IntelliJ files: https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore

evpaassen:
@pathikrit We follow the recommendations by Jetbrains: https://www.jetbrains.com/idea/help/managing-projects-under-version-control.html

If we wouldn't share these settings, it would be a huge pain to configure code style settings, etc. in the IDE.

Issue #169, opened 2015-01-07
Unable to bind to recurring appointments due to date formatting


ahaarrestad:
The date-format for the recurrence block for some of our legacy appointments have for format like 2014-06-11-5:00. The code today only sypports 2014-06-11Z.

vboctor:
@ahaarrestad is this date parsing an issue across the board or for this specific scenario?  Is your fix going to address all similar APIs?  I haven't looked at the pull request yet.  The goal would be to have consistency on handled date formats.

ahaarrestad:
The goal is to handle the same date-formats for recurring events as for any events. I created the pull request because of the recurring events, and that's what it fixes. But it also cleans up some of the code for parsing dates.

PR #175, opened 2015-01-15
Improve parsing of dates for recurring events

ahaarrestad:
See https://github.com/OfficeDev/ews-java-api/pull/165 and https://github.com/OfficeDev/ews-java-api/issues/169 for discussion.

PR #176, opened 2015-01-15
Rework date parsing

evpaassen:
This addresses #169 and reworks date parsing.
Alternative implementation of #175.


azurecla:

TTYL, AZPRBOT;

ahaarrestad:
I would suggest much more unit-tests when shortening the code like this.

evpaassen:
But the tests cover almost all the cases. I also made sure all cases from the original method are supported. The main thing is that we don't need to parse the timezone suffix manually as this is already supported by SimpleDateFormat.

This code is actually close to the original, and doesn't refactor between methods.

serious6:
have to admit I like the shortened version most. @evpaassen for future PR I would appreciate if you add comments and thoughts to a given PR before we break down to competitions. Since both of these PR reference the same issue and there is only a small difference between them.

ahaarrestad:
there is one huge difference... https://github.com/OfficeDev/ews-java-api/pull/176 does not fix issue https://github.com/OfficeDev/ews-java-api/issues/169 since it does not implement fixes for the recurring appointments.

src/test/java/microsoft/exchange/webservices/data/ExchangeServiceBaseDateParsingTest.java
```diff
@@ -0,0 +1,185 @@
+package microsoft.exchange.webservices.data;
```
serious6: ExchangeServiceBaseDateParsingTest is missing license-header

serious6:
I leave it up to you how you handle this. Maybe another solution could be if @evpaassen creates a PR on your PR @ahaarrestad.

evpaassen:
@serious6 Let's have this clear. This is not about competition, or credits or anything. Sometimes it's just easier to explain a different concept with an example. I'm just trying to be helpful here.

@ahaarrestad The tests are of course heavily inspired by your tests, as those were looking quite nice. I took one test from yours and adjusted it to verify every pattern supported by the old and new parsing code.

> there is one huge difference... #176 does not fix issue #169 since it does not implement fixes for the recurring appointments.

Doesn't it? It should be fixed by f7a7c52. `testStartDateWithTimeZone()` and `testStartDateWithTimeZoneWithColon()` seem to verify it's fixed, not?

src/test/java/microsoft/exchange/webservices/data/ExchangeServiceBaseDateParsingTest.java
evpaassen: The license header has been added.

ahaarrestad:
Your fix won't handle the actual problem that occurs if the recurring appointment contains different timezones in the string. Thus my change from the old parsing to the reworked one.

ahaarrestad:
@serious6 your call on how to continue https://github.com/OfficeDev/ews-java-api/pull/175 vs https://github.com/OfficeDev/ews-java-api/pull/176

evpaassen:
OK, I reworked it a little bit again. I guess I misunderstood the original issue a little bit. It should fix #169 now, though.

It contains an additional fix for date-only parsing when the date string has no timezone. The library should then assume UTC instead of local timezone. This bug was fixed for date-time string already in #6.

serious6:
@evpaassen, @ahaarrestad  hi I aprreciate both of your contributions and encourage on this topic. Thanks for that in advance. 

For solving this kind of problem we ran in yesterday maybe @ahaarrestad can provide a test which underlines the expected behavior for the recurring appointment problem, since he pointed out that this PR does not include the fix for this problem. After the integration is done or disproved maybe we can have a merge on @evpaassen PR because it contains the "easier"-implementation for the parsing methods. I dont think it is usefull to merge both of the PR. What do you think of this?

ahaarrestad:
I'll close my pr with refs to this one.

serious6:
@evpaassen are you finished with this? Thanks for the refactoring. I think it´s even better now.

evpaassen:
Yes, I'm finished now.

serious6:
**Thanks for the contribution @evpaassen , @ahaarrestad**
_Merged with 21c7b0dacc9fb7ac5eb65fe8ee24adb544f9f52e_

evpaassen:
@ahaarrestad Thank you very much for your collaboration on fixing this issue. I really appreciate it.

src/main/java/microsoft/exchange/webservices/data/util/DateTimeParser.java
```diff
@@ -0,0 +1,101 @@
...
+      for (SimpleDateFormat format : formats) {
+        try {
+          return format.parse(value);
+        } catch (ParseException e) {
+          // Ignore and try the next pattern.
+        }
+      }
+    }
+
+    throw new IllegalArgumentException(
```
vboctor: Is there a reason for throwing a different kind of exception here?  IllegalArgumentException vs. ParseException?  I generally prefer to use illegal argument kind of exceptions to reflect programming errors where exceptions based on runtime data would use ParseException (or FormatException in .NET).  Thoughts?

src/main/java/microsoft/exchange/webservices/data/util/DateTimeParser.java
```diff
@@ -0,0 +1,101 @@
...
+ **************************************************************************/
+
+package microsoft.exchange.webservices.data.util;
+
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.TimeZone;
+
+public class DateTimeParser {
```
vboctor: Do we want to expose this for consumers of the library?  Should we make it package internal?

src/main/java/microsoft/exchange/webservices/data/UserConfigurationDictionary.java
```diff
@@ -581,8 +583,7 @@ private Object constructObject(UserConfigurationDictionaryObjectType type,
     } else if (type.equals(UserConfigurationDictionaryObjectType.ByteArray)) {
       dictionaryObject = Base64EncoderStream.decode(value.get(0));
     } else if (type.equals(UserConfigurationDictionaryObjectType.DateTime)) {
-      Date dateTime = reader.getService()
-          .convertUniversalDateTimeStringToDate(value.get(0));
+      Date dateTime = new DateTimeParser().convertDateTimeStringToDate(value.get(0));
       if (dateTime != null) {
```
vboctor: What does dateTime == null mean? Doesn't the underlying class throw exceptions if the date is not valid?

src/test/java/microsoft/exchange/webservices/data/util/DateTimeParserTest.java
```diff
@@ -0,0 +1,219 @@
...
+
+import java.util.Calendar;
+import java.util.Date;
+import java.util.GregorianCalendar;
+import java.util.TimeZone;
+
+import static org.junit.Assert.assertEquals;
+
+@RunWith(JUnit4.class)
+public class DateTimeParserTest {
```
vboctor: Should we add test cases for invalid format and null cases?

vboctor:
@evpaassen and @ahaarrestad great to see us fixing issues as well as making the API and testability better.  I was a bit late to the party, but I've added a couple of comments anyway that we can possibly address as a follow up.

src/main/java/microsoft/exchange/webservices/data/util/DateTimeParser.java
evpaassen: I placed this class in a subpackage (yes, the first subpackage in this library), so we can't make it package private if we keep it there. This shouldn't be a problem though. All major libraries or frameworks expose internal utility classes like this. Otherwise package organisation wouldn't be possible.

src/test/java/microsoft/exchange/webservices/data/util/DateTimeParserTest.java
evpaassen: You're right, we should *at least* have a test case with null and the empty string. I'll add them when I have some time.

src/main/java/microsoft/exchange/webservices/data/util/DateTimeParser.java
evpaassen: The former method `convertUniversalDateTimeStringToDate(...)` already did so, so I kept this behaviour. Note that IllegalArgumentException is a runtime exception, where ParseException is a checked exception. I think that most important is that this behaviour is documented in the javadoc, like it is.

src/test/java/microsoft/exchange/webservices/data/util/DateTimeParserTest.java
vboctor: We should also test the case of throwing exceptions.

src/main/java/microsoft/exchange/webservices/data/util/DateTimeParser.java
vboctor: So if we change a utility method interface, we will consider this a contract change?  Hence, prefer to mark them as deprecated unless we are working on a major version?  Or should we have SomeNameSpace.Internal.* to reflect that these are internal packages that shouldn't be used by client apps?

src/main/java/microsoft/exchange/webservices/data/util/DateTimeParser.java
vboctor: Yep, I've just read about [checked vs. runtime exceptions](http://johnpwood.net/2008/04/21/java-checked-exceptions-vs-runtime-exceptions/), I'm new to the concept of checked exceptions since it is unique to Java.  However, I would have thought that parsing should be a checked exception to force the callers to handle the case where the string they are providing is not valid.  This is not something that you want to bubble to the top of the stack.  It is interesting however that SimpleDateFormat throws ParseException (checked) and long.ParseLong throws IllegalArgumentException (runtime).  I would have thought there would be consistency there and both use checked, since I would expect parsing errors to be handled closest to where they are used.

PR #177, opened 2015-01-16
now using the osgi enabled dependency replacement for non osgi libraries

Arcticwolf:
This change needs a proper test if everything work as expected, since I'm not sure if the new libraries behave the same like the previously used ones.

### DIFF
--- a/.idea/libraries/Maven__org_apache_httpcomponents_fluent_hc_4_3_5.xml
+++ b/.idea/libraries/Maven__org_apache_httpcomponents_fluent_hc_4_3_5.xml
@@ -1,0 +1,13 @@
+<component name="libraryTable">
+  <library name="Maven: org.apache.httpcomponents:fluent-hc:4.3.5">
+    <CLASSES>
+      <root url="jar://$MAVEN_REPOSITORY$/org/apache/httpcomponents/fluent-hc/4.3.5/fluent-hc-4.3.5.jar!/" />
+    </CLASSES>
+    <JAVADOC>
+      <root url="jar://$MAVEN_REPOSITORY$/org/apache/httpcomponents/fluent-hc/4.3.5/fluent-hc-4.3.5-javadoc.jar!/" />
+    </JAVADOC>
+    <SOURCES>
+      <root url="jar://$MAVEN_REPOSITORY$/org/apache/httpcomponents/fluent-hc/4.3.5/fluent-hc-4.3.5-sources.jar!/" />
+    </SOURCES>
+  </library>
+</component>
--- a/.idea/libraries/Maven__org_apache_httpcomponents_fluent_hc_4_3_6.xml
+++ b/.idea/libraries/Maven__org_apache_httpcomponents_fluent_hc_4_3_6.xml
@@ -1,0 +1,13 @@
+<component name="libraryTable">
+  <library name="Maven: org.apache.httpcomponents:fluent-hc:4.3.6">
+    <CLASSES>
+      <root url="jar://$MAVEN_REPOSITORY$/org/apache/httpcomponents/fluent-hc/4.3.6/fluent-hc-4.3.6.jar!/" />
+    </CLASSES>
+    <JAVADOC>
+      <root url="jar://$MAVEN_REPOSITORY$/org/apache/httpcomponents/fluent-hc/4.3.6/fluent-hc-4.3.6-javadoc.jar!/" />
+    </JAVADOC>
+    <SOURCES>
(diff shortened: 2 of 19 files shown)


azurecla:
Hi __@Arcticwolf__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

azurecla:
__@Arcticwolf__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

MikeN123:
I'm wondering if we really want to make this library OSGi ready? Larger projects (e.g., Spring) have dropped OSGi support quite a while ago, and I am not too keen on all kinds of 'OSGi ready' dependencies in my non-OSGi projects.

serious6:
I am also no friend of the applied changes since they will result in keeping an eye on many files after libraries changed. To underline @MikeN123 post here is a [stackoverflow](http://stackoverflow.com/a/21868745) for this topic.

Arcticwolf:
Of course this decision is not mine : ) If you don't want to have the library OSGi enabled, then this is ok too. 
Just my two cents about OSGi: IMHO it is not true that larger projects in general go away from OSGi (e.g. almost everything in the Apache environment or the whole Eclipse universe). I know that Spring went away from OSGi but the other way round the OSGi world went step by step away from Spring, since Apache Blueprint is getting more and more a replacement of Spring which worked way more smoothly with OSGi. Naturally Blueprint went way better with OSGi, since it was designed with OSGi in mind : )

MikeN123:
In my opinion, the only feasible solution would be a separate artifact, using a separate maven profile with the OSGi dependencies (or something like that, I'm no Maven expert). That way, the OSGi-specific dependencies can stay out of the build for non-OSGi users. It is also the way HttpClient handles it, apparently they use a separate OSGi project: https://svn.apache.org/repos/asf/httpcomponents/httpclient/tags/4.3.6/httpclient-osgi/

vboctor:
Looks like we have two -1s for this.  I like the idea of ews-java-api-osgi that is published from a fork whenever an ews-java-api released is published, but I don't have a lot of experience around this.

@evpaassen what's your take on this one?

evpaassen:
> Looks like we have two -1s for this. I like the idea of ews-java-api-osgi that is published from a fork whenever an ews-java-api released is published, but I don't have a lot of experience around this.

I never used OSGi, but I'm not keen on having OSGi-specific dependencies (like `httpclient-osgi`) in this library when not using OSGi. So I think there should be a OSGi-enabled fork indeed, which could be maintained by someone who is actually using OSGi or has sufficient experience with it.

I'd suggest backing out #98 for now.

vboctor:
I've reverted #98.  Will also close this PR.  @Arcticwolf I assume having a fork on github that builds OSGi compliant versions would address your needs.  However, let's keep any core improvements in the main repositories, and use that just for mixing in OSGi.  Thanks everyone.

tomravn:
We need an OSGi-ready version of this library. For now we've forked and done it ourselves, but given there was already a PR for this I imagine it will be requested in the future too. Given the changes were made, it seems a shame that they were reverted without a branch being made first!

The only thing required to make a project OSGi-ready is some extra entries in the manifest. In the case of httpclient, the artifact needs to change but as far as I know the osgi and non-osgi versions are interchangeable. 99% of projects don't have different artifacts for OSGi and non-OSGi and I can't see a reason why this project needs to keep them separate. Making something OSGi-ready does not stop it working in non-OSGi environments.

Once you've added a (quite simple) maven plugin, all the manifest entries are handled automatically so there's no extra files to maintain etc. The only other changes required are the httpcllient artifact names and the packaging type from jar to bundle. I've made and tested these changes on our fork. I can create a PR if you want

MikeN123:
> Making something OSGi-ready does not stop it working in non-OSGi environments.

It does, since we have more dependencies that require HttpClient, and thus I end up with multiple HttpClient versions on my class path.

tomravn:
OSGi or not, unless those other dependencies use exactly the same version of http client as this project, you should exclude one of the http clients to be sure about which one maven will put on your classpath. If this project was made osgi ready, then you might have to change the artifact that's excluded but that's the only difference.

NB, having many dependencies which each want a different version of httpclient was one of our main reasons for moving to OSGi.

MikeN123:
Uhm, I'm not sure about your Maven, but our Gradle (and Maven too IIRC) do dependency version resolution, and simply pick the highest required version.

Issue #100, opened 2014-10-16
HTTP connection management


evpaassen:
Currently, the HttpConnectionManager used by `ExchangeServiceBase` is not configurable. It would be useful to have the connection manager as an optional constructor argument, so it can be configured properly.

I'd be happy to submit a pull request.

evpaassen:
The shutdown issue has been resolved and the connection manager has been replaced by a basic one in #180. So this issue can be closed now.

Issue #162, opened 2015-01-06
Where ExchangeServiceBase.preAuthenticate is used?


barisaydinoglu:
I think this field comes from C# code, and could not found it's usage.

serious6:
it´s being set but never used with the `HttpWebRequest`

vboctor:
@evpaassen's last change touched on this.  I also just sent him a question about whether there is a reason why we wouldn't drop it since it is dead code.

evpaassen:
In #22, I enabled preemptive authentication, but this has been removed (unintendedly) when migrating to HttpClient 4 in #113, #126. My implementation in #22 should have used this `preAuthenticate` field to disable preemptive authentication when performing a plain HTTP autodiscover request, but it didn't.

So currently, it is dead code indeed. But preemptive authentication should be re-implemented in the correct way and the `preAuthenticate` field of `HttpWebRequest` should be used to determine whether preemptive auth should be enabled for that single request.

It's not a big deal, but it would save some overhead. I could try to reimplement it correctly when I have some time. I'd suggest leaving this code in place until then. I already added comments about this on the getter and setter of `preAuthenticate` in `HttpWebRequest`.

PR #180, opened 2015-01-19
Rework connection management, reuse HttpClient

evpaassen:
This pull request is primary meant to fix the `HttpConnectionManager` shutdown issue mentioned in #100. It adds a `close()` method to `ExchangeServiceBase`, which shuts down the `HttpClient`, which in turn shuts down the `HttpConnectionManager`. This required refactoring, because the `HttpClient` wasn't accessible from `ExchangeServiceBase`, only from the `HttpWebRequest` as every request created a new client. The `HttpClient` should be reused anyway for performance reasons, so it has been moved to the `ExchangeServiceBase`.

Other (related) changes:
* The `PoolingHttpClientConnectionManager` is replaced by `BasicHttpClientConnectionManager`. As `ExchangeServiceBase` is not thread-safe anyway, there's no need to use a connection pool: a basic connection manager does suffice.
* The library touched the global `CookieHandler` settings for no reason. This has been fixed.
* The JCIFS dependency has been removed, because it is unused since the migration to HttpClient 4 (#126).
* The `AutodiscoverService` did not close a response object on early return in `getLegacyUserSettingsAtUrl(...)`. The response object is now always being closed properly.
* The scope of the proxy URL was per-request, while proxy-credentials were global (allowing ony one set of proxy credentials, globally). The HttpProxyCredentials has been replaced by WebProxyCredentials, which have a per-request scope, just like the URL.
* Cookies are now handled implicitly by the HttpClient, instead of explicitly by the ourselves.

*Note: please don't squash these commits, as they fix separate (though related) issues.*


azurecla:

TTYL, AZPRBOT;

MikeN123:
I added a commit that only allows NTLM and Basic authentication, to prevent the Kerberos warnings/slowness mentioned in #144 and #160. Basic authentication is necessary since Office 365 only seems to support Basic auth.

src/main/java/microsoft/exchange/webservices/data/HttpWebRequest.java
```diff
@@ -220,18 +213,14 @@ public void setUrl(URL url) {
   }
 
   /**
-   * Checks if is pre authenticate.
-   *
-   * @return true, if is pre authenticate
+   * Whether to use preemptive authentication. Currently not implemented, though.
```
vboctor: Any benefit of having this while it doesn't have an effect?  I can see we have other public interface changes here, so it is not like we don't want to change the public interface.

vboctor:
Thanks @evpaassen and @MikeN123.  Merged.

src/main/java/microsoft/exchange/webservices/data/HttpWebRequest.java
evpaassen: See https://github.com/OfficeDev/ews-java-api/issues/162#issuecomment-71319928.

PR #187, opened 2015-01-26
Deploy artifacts to maven central

serious6:
This PR is about Issue: #1. But also adds inprovements for building a local Jar-File.
We are able to create Jars easily now `mvn clean install` provides 3 jars now (`PROJECT_ROOT\target`):
* ews-java-api-1.3-SNAPSHOT.jar
* ews-java-api-1.3-SNAPSHOT-javadoc.jar
* ews-java-api-1.3-SNAPSHOT-sources.jar

I also added the ability for deploying to sonatype by integrating `nexus-staging-maven-plugin`.
`mvn clean deploy` will now publish jars to sonatype if valid credentials are stored in settings.xml (just tested until the upload cause I dont have the credentials). If the release-name end with snapshot it will automaticly been stored to snapshot repo.

settings.xml (`~/.m2/settings.xml`) should look like this:
```xml
<settings>
  ...
  <servers>
    ...
    <server>
      <id>ossrh</id> <!-- ref to id in pom.xml / distributionManagement --> 
      <username>Jira username</username>
      <password>Jira password</password>
    </server>
  </servers>
  ...
</settings>
```
The password should be stored [encrypted](http://maven.apache.org/guides/mini/guide-encryption.html).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
-    <groupId>com.microsoft.office</groupId>
+    <groupId>com.microsoft.ewsjavaapi</groupId>
     <artifactId>ews-java-api</artifactId>
     <version>1.3-SNAPSHOT</version>
 
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,7 @@
     <groupId>com.microsoft.ewsjavaapi</groupId>
     <artifactId>ews-java-api</artifactId>
     
-    <version>1.3-SNAPSHOT</version>
+    <version>2.0-SNAPSHOT</version>
 
     <name>Exchange Web Services Java API</name>
     <description>Exchange Web Services (EWS) Java API</description>
--- a/pom.xml
+++ b/pom.xml
@@ -6,6 +6,7 @@
 
     <groupId>com.microsoft.ewsjavaapi</groupId>
     <artifactId>ews-java-api</artifactId>
+    
     <version>1.3-SNAPSHOT</version>
 
     <name>Exchange Web Services Java API</name>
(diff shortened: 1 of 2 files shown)


azurecla:

TTYL, AZPRBOT;

pom.xml
```diff
@@ -77,6 +113,17 @@
...
     <build>
         <plugins>
+            <!--Deployment / build plugins-->
+            <plugin>
+                <groupId>org.sonatype.plugins</groupId>
+                <artifactId>nexus-staging-maven-plugin</artifactId>
+                <version>1.6.5</version>
+                <extensions>true</extensions>
+                <configuration>
+                    <!--Ref.: http://books.sonatype.com/nexus-book/reference/staging-deployment.html-->
```
ahaarrestad: this line looks a bit strange ..... is it correct?

serious6: you are referring to the comment?

serious6: @ahaarrestad I have seen that there was a blank between <!-- and the comment message in the other comments. Maybe this was the confusion. Added the missing blank.

pom.xml
```diff
@@ -4,8 +4,9 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
-    <groupId>com.microsoft.office</groupId>
+    <groupId>com.microsoft.ewsjavaapi</groupId>
     <artifactId>ews-java-api</artifactId>
+    
     <version>1.3-SNAPSHOT</version>
```
vboctor: I suspect that by now we have broken the public API compared to 1.2, I wonder if we should move the version to 2.0-SNAPSHOT?  Is there a tool that compares the 1.2 vs. 1.3 and tell us if we have broken the library public interface?  I suspect we will break it before release.  Thoughts?

vboctor:
Thanks @serious6, it is great to have this.  The PR itself has very useful information, should we include this information in the readme.md file?  Otherwise, looks good.

serious6:
yes. We should have this updated too. I will think of a text that is suitable.

pom.xml
serious6: hmm never heard of a tool which would give us that information. Maybe this is ~~"just"~~ about looking through the git-log? But I appreciate the idea of deploying `2.0-Snapshot` releases now since we had some greater changes and there will possibly join in more of these until we are releasing the first stable one.

pathikrit:
When do we plan to deploy the artifacts into maven? We should probably also put the Maven groupId etc or put a link to something like this in the README:https://www.versioneye.com/java/com.microsoft.windowsazure:microsoft-windowsazure-api

vboctor:
Looks good, let's move the version to 2.0 as well.  This will give us more flexibility to do a good initial cleanup.

serious6:
version is updated to `2.0-SNAPSHOT` now.

serious6:
@pathikrit we will update the information you mentioned after @vboctor did  the deployment.

Issue #181, opened 2015-01-20
SimpleDateFormat doesn't support timezone parsing with the character X in the pattern string


derk5008:
I downloaded latest code and it build successfully. But when I try to connect to EWS, I still got error:

Jan 20, 2015 2:17:58 PM org.apache.http.impl.auth.HttpAuthenticator generateAuthResponse
WARNING: NEGOTIATE authentication error: Invalid name provided (Mechanism level: Could not load configuration file C:\WINDOWS\krb5.ini (The system cannot find the file specified))
Exception in thread "main" microsoft.exchange.webservices.data.ServiceRequestException: The request failed. The request failed. The remote server returned an error: (401)Unauthorized

The code I am using is from "Download ZIP". I have httpclient-4.3.5, httpcore-4.3.3 in build path. And here is my client code: 
ExchangeService service = new ExchangeService(); 
		ExchangeCredentials credentials = new WebCredentials("username", "password", "myDomain"); 
		service.setCredentials(credentials);
		service.setUrl(new URI("https://ewsURL"));  
		ItemView view = new ItemView(5);  
		FindItemsResults<Item> findResults = service.findItems(WellKnownFolderName.Inbox, view);	

Could anyone give some suggestion? I really appreciate it! Thanks a lot in advance!

avromf:
Hi,

In the past, I have found the following website useful when trying to test connections to EWS.
https://testconnectivity.microsoft.com/
Select "Service Account Access (Developers)" under "Microsoft Exchange Web Services Connectivity Tests".
This lets you know whether it is a general connection problem vs. a EWS Java API problem.

derk5008:
Hey, avromf, thank you very much! The Test couldn't go through, but I think the reason could be the firewall: my EWS is behind the firewall so it can't be reached from outside....But thanks a lot for the tips. It gives me one more way to test.

derk5008:
OK...Got some update. I rebuild the code and my project, then the precious error is gone. But now I have following error:

Jan 20, 2015 3:18:34 PM org.apache.http.impl.auth.HttpAuthenticator generateAuthResponse
WARNING: NEGOTIATE authentication error: Invalid name provided (Mechanism level: Could not load configuration file C:\WINDOWS\krb5.ini (The system cannot find the file specified))
Exception in thread "main" microsoft.exchange.webservices.data.ServiceRequestException: The request failed. Illegal pattern character 'X'

[...]

PR #189, opened 2015-01-27
Building against multiple jdk

serious6:
As mentioned in #181 we ran into issues because there where differences in the used jdk. This PR tries to prevent this in the future by adding maven profiles for travis-ci which will now build the api against all available jdk's. This also prevents the build from failing while compiling with jdk1.8 with enabled doclint function.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,12 @@
 language: java
 
+jdk:
+  - oraclejdk8
+  - oraclejdk7
+  - openjdk6
+
 notifications:
   webhooks:
     urls:
       - https://webhooks.gitter.im/e/04643d955d03eb39e24e
-    on_success: change  # options: [always|never|change] default: always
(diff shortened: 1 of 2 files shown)


azurecla:

TTYL, AZPRBOT;

vboctor:
Will this make it a requirement that every developer has to have all supported JDKs installed?

serious6:
No it is just about the Travis build. Local building remains uneffected.

evpaassen:
I think we should not use seperate build profiles for CI. The build should from CI should be reproducable by the developer. Ideally, the `pom.xml` should not know anything about CI.

Furthermore, building with JDK 1.7 or JDK 1.8 does not mean we need to adjust the Java source and target level. In fact, this should always be `1.6`.

serious6:
hmm maybe you are right. This indicates also to add (for the programmer unused) overhead to the `pom.xml`. I will update the PR.

PR #194, opened 2015-02-03
Removed Abstract from ILazyMember interface. It is not needed, and for n...

ahaarrestad:
...ewer java-versions it seems to clutter up the compiling a bit.


serious6:
@ahaarrestad **thanks for your contribution**

PR #196, opened 2015-02-03
fix: Issue 28 and add improvement for response parameters

serious6:
This PR will fix issue 28 which outlined that the response-object from an internalExecute will be closed twice which will result in an error. I also tried to add some improvement to the handling of return-parameters as well since they where just defined as Object and dont include reflection. 

This should not be merged since the functionality of the methods changed is confirmed.
Therefore help with testing would be welcome as well as your comments and suggestions.

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/ServiceRequestBase.java
+++ b/src/main/java/microsoft/exchange/webservices/data/ServiceRequestBase.java
@@ -23,17 +23,15 @@
 package microsoft.exchange.webservices.data;
 
 import javax.xml.stream.XMLStreamException;
-import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
-import java.io.InputStream;
+import javax.xml.ws.http.HTTPException;
+import java.io.*;
 import java.util.zip.GZIPInputStream;
 import java.util.zip.InflaterInputStream;
 
 /**
  * Represents an abstract service request.
  */
-abstract class ServiceRequestBase {
+abstract class ServiceRequestBase<T> {
 
   // Private Constants
   // private final String XMLSchemaNamespace =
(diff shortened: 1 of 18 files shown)


azurecla:

TTYL, AZPRBOT;

src/main/java/microsoft/exchange/webservices/data/ServiceRequestBase.java
```diff
@@ -368,13 +356,85 @@ private static InputStream getResponseErrorStream(HttpWebRequest request)
...
+    try {
+      this.getService().processHttpResponseHeaders(
+          TraceFlags.EwsResponseHttpHeaders, response);
+
+      if (this.getService().isTraceEnabledFor(TraceFlags.EwsResponse)) {
+        ByteArrayOutputStream memoryStream = new ByteArrayOutputStream();
+        InputStream serviceResponseStream = ServiceRequestBase
+            .getResponseStream(response);
+        while (true) {
+          int data = serviceResponseStream.read();
```
ahaarrestad: This can be squashed in to the while-loop to avoid the "white(true)" stuff.....

ahaarrestad:
Any chance of adding tests for at least parts of the changed code?

src/main/java/microsoft/exchange/webservices/data/ServiceRequestBase.java
serious6: interchanged the while(true) it was previously just copy paste from the original one.

serious6:
since the changed code is all about request and response the underlying code can just be tested with a given exchange-server and account information. I will try to add some of them in the next couple of days to initially handle that kind of tests.

serious6:
we need some test-accounts to run a suite against. Pinged @vboctor about this.

serious6:
automated tests will be added with a seperate PR

PR #201, opened 2015-02-05
Properly close request when HttpErrorException is encountered

evpaassen:
The request isn't returned and can't be closed by the caller in case of an `HttpErrorException`.

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/ServiceRequestBase.java
+++ b/src/main/java/microsoft/exchange/webservices/data/ServiceRequestBase.java
@@ -621,13 +621,24 @@ abstract class ServiceRequestBase {
     this.validate();
 
     HttpWebRequest request = this.buildEwsHttpWebRequest();
-    try {
-      return this.getEwsHttpWebResponse(request);
-    } catch (HttpErrorException e) {
-      processWebException(e, request);
 
-      // Wrap exception if the above code block didn't throw
-      throw new ServiceRequestException(String.format(Strings.ServiceRequestFailed, e.getMessage()), e);
+    try {
+      try {
+        return this.getEwsHttpWebResponse(request);
+      } catch (HttpErrorException e) {
+        processWebException(e, request);
+
+        // Wrap exception if the above code block didn't throw
+        throw new ServiceRequestException(String.format(Strings.ServiceRequestFailed, e.getMessage()), e);
+      }
+    } catch (Exception e) {
+      try {
+        request.close();
+      } catch (Exception e2) {
+        // Ignore exceptions while closing the request.
+      }
+
+      throw e;
     }
   }
 


azurecla:

TTYL, AZPRBOT;

src/main/java/microsoft/exchange/webservices/data/ServiceRequestBase.java
```diff
@@ -624,7 +624,11 @@ protected HttpWebRequest validateAndEmitRequest() throws ServiceLocalException,
     try {
       return this.getEwsHttpWebResponse(request);
     } catch (HttpErrorException e) {
-      processWebException(e, request);
+      try {
+        processWebException(e, request);
+      } finally {
```
ahaarrestad: should the request be closed after return? meaning moving the finally to the outer catch-block?

evpaassen: No, only if an exception occurs.

src/main/java/microsoft/exchange/webservices/data/ServiceRequestBase.java
```diff
@@ -621,13 +621,24 @@ protected HttpWebRequest validateAndEmitRequest() throws ServiceLocalException,
...
+        return this.getEwsHttpWebResponse(request);
+      } catch (HttpErrorException e) {
+        processWebException(e, request);
 
-      // Wrap exception if the above code block didn't throw
-      throw new ServiceRequestException(String.format(Strings.ServiceRequestFailed, e.getMessage()), e);
+        // Wrap exception if the above code block didn't throw
+        throw new ServiceRequestException(String.format(Strings.ServiceRequestFailed, e.getMessage()), e);
+      }
+    } catch (Exception e) {
```
vboctor: Why does this catch block wrap the other one rather than be part of the same try/catch block?

vboctor:
+1

src/main/java/microsoft/exchange/webservices/data/ServiceRequestBase.java
evpaassen: Because the request should be closed on every Exception (because the request is not returned in that case), including the ones raised in `processWebException()`. The exception handling in this libary is very messy, unfortunately.

ahaarrestad: so the request should stay open after returm?

MikeN123: Yeah, it should. In all non-error cases it is properly closed, but in case of an error it is not returned, so cannot be closed by the caller. It's a bit weird, the whole error handling could probably use some refactoring in the future.

ahaarrestad:
+1

ahaarrestad:
time to merge this one?

PR #202, opened 2015-02-05
Keep whitespace in plain text email messagebodys

### LINKED ISSUE #198: Whitespaces removed from plain text email messagebody
Hi,

my original - plain text - email looks like this:
```
Line 1
Line 2, followed by empty 3. line

Line 4, followed by empty 5. and 6. line


 Line 7 with one leading space
  Line 8 with two leading spaces
> line 9 with leading ">"
 > line 10 with leading space + ">"
  > line 11 with 2 leading spaces + ">", followed by 2 empty lines  


Line 14 (last line)
```

What is received by ews-java-api - with  BodyType.Text - is:
```
Line 1
Line 2, followed by empty 3. line
Line 4, followed by empty 5. and 6. line
 Line 7 with one leading space
  Line 8 with two leading spaces> line 9 with leading ">" > line 10 with leading space + ">"  > line 11 with 2 leading spaces + ">", followed by 2 empty lines  
Line 14 (last line)
```

After some investigation I found the problem is in MessageBody#readTextValueFromXml() and EwsXmlReader#read()  were "unwanted" whitespace is always removed.
I made some small changes in these two files to keep whitespaces for plain text messagebodys.
Now plain text messages come out as expected.

If you are interested in my solution I can provide a patch file or make a fork & pull request.
Just let me know.

superelchi

superelchi:
Fixes #198

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/EwsXmlReader.java
+++ b/src/main/java/microsoft/exchange/webservices/data/EwsXmlReader.java
@@ -152,6 +152,8 @@ class EwsXmlReader {
 
   /**
    * Reads the specified node type.
+   * 
+   * @param keepWhiteSpace Do not remove whitespace characters if true
    *
    * @throws ServiceXmlDeserializationException  the service xml deserialization exception
    * @throws javax.xml.stream.XMLStreamException the xML stream exception
@@ -408,6 +410,7 @@ class EwsXmlReader {
    * Present event must be START ELEMENT. After executing this function
    * Present event will be set on END ELEMENT
    *
+   * @param keepWhiteSpace Do not remove whitespace characters if true
    * @return String
    * @throws javax.xml.stream.XMLStreamException the xML stream exception
    * @throws ServiceXmlDeserializationException  the service xml deserialization exception
--- a/src/main/java/microsoft/exchange/webservices/data/EwsXmlReader.java
+++ b/src/main/java/microsoft/exchange/webservices/data/EwsXmlReader.java
@@ -147,6 +147,17 @@ class EwsXmlReader {
    */
   public void read() throws ServiceXmlDeserializationException,
       XMLStreamException {
+    read(false);
+  }
+
+  /**
+   * Reads the specified node type.
+   *
+   * @throws ServiceXmlDeserializationException  the service xml deserialization exception
+   * @throws javax.xml.stream.XMLStreamException the xML stream exception
+   */
+  private void read(boolean keepWhiteSpace) throws ServiceXmlDeserializationException,
+      XMLStreamException {
     // The caller to EwsXmlReader.Read expects
     // that there's another node to
     // read. Throw an exception if not true.
@@ -158,9 +169,10 @@ class EwsXmlReader {
         XMLEvent event = xmlReader.nextEvent();
         if (event.getEventType() == XMLStreamConstants.CHARACTERS) {
           Characters characters = (Characters) event;
-          if (characters.isIgnorableWhiteSpace()
-              || characters.isWhiteSpace()) {
-            continue;
+          if (!keepWhiteSpace)
+            if (characters.isIgnorableWhiteSpace()
+                || characters.isWhiteSpace()) {
+              continue;
           }
         }
         this.prevEvent = this.presentEvent;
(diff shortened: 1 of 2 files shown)


azurecla:
Hi __@superelchi__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

azurecla:
__@superelchi__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

ahaarrestad:
Please add a unit-test to test the code you have changed :-)

src/main/java/microsoft/exchange/webservices/data/EwsXmlReader.java
```diff
@@ -147,6 +147,17 @@ private void internalReadElement(String namespacePrefix, String localName,
    */
   public void read() throws ServiceXmlDeserializationException,
       XMLStreamException {
+    read(false);
```
ahaarrestad: Please correct the intendation.

src/main/java/microsoft/exchange/webservices/data/MessageBody.java
```diff
@@ -100,7 +100,8 @@ protected void readAttributesFromXml(EwsServiceXmlReader reader)
   @Override
   protected void readTextValueFromXml(EwsServiceXmlReader reader)
       throws XMLStreamException, ServiceXmlDeserializationException {
-    this.text = reader.readValue();
+    boolean keepWhiteSpace = this.getBodyType() == BodyType.Text;
```
ahaarrestad: no need to store this variable. put it directly into the method below. Also, it is safer to use equals instead of ==

evpaassen: `BodyType` is an enum, so there's absolutely no need to compare using the equals method. `==` is just fine here.

src/main/java/microsoft/exchange/webservices/data/EwsXmlReader.java
evpaassen: The indentation is correct.

superelchi:
If you provide the basic unit-tests for testing MessageBody and / or EwsXmlReader I will gladly add tests for my changes. ;)

ahaarrestad:
you can test readTextValueFromXml(...), and there is no need for private methods. If you don't write private/protected you get the in-between security :-) and yoiu can test it

superelchi:
I do nor understand - how should I write a "standalone" test for readTextValueFromXml?
It need a EwsServiceXmlReader, connected to a Soap-message, ...
So the test would be more than 10 times the size and complexity the real change is.
Or am I wrong?

src/main/java/microsoft/exchange/webservices/data/EwsXmlReader.java
ahaarrestad: my bad. just reading the diff wrong

src/main/java/microsoft/exchange/webservices/data/MessageBody.java
ahaarrestad: good point

ahaarrestad:
+1

src/main/java/microsoft/exchange/webservices/data/EwsXmlReader.java
```diff
@@ -158,9 +171,10 @@ public void read() throws ServiceXmlDeserializationException,
         XMLEvent event = xmlReader.nextEvent();
```
serious6: maybe we can also get rid of this scary `while(true)` **thing**?

src/main/java/microsoft/exchange/webservices/data/MessageBody.java
```diff
@@ -100,7 +105,15 @@ protected void readAttributesFromXml(EwsServiceXmlReader reader)
   @Override
   protected void readTextValueFromXml(EwsServiceXmlReader reader)
       throws XMLStreamException, ServiceXmlDeserializationException {
-    this.text = reader.readValue();
+    if (log.isDebugEnabled()) {
+       log.debug("Reading text value from XML. BodyType = " + this.getBodyType() +
+          ", keepWhiteSpace = " +
+          ((this.getBodyType() == BodyType.Text) ? "true." : "false."));
```
serious6: please use String.format(...) for the combination of strings. (same @ line 115)

superelchi: I intentionally used String concat here because
- it's about 10 times faster than String.format()
- it's used in other ews-java-api log messages already, see SimpleServiceRequestBase.java:123:
```
      log.error("Response content type not XML; first line: '" + line + "'");
```
So I will not change it.

serious6:
@superelchi thanks for the contribution. Once you'll have the comments addressed this can be merged.

superelchi:
From my point of view all comments are already handled.

The sacry "while(true) .." does nor belong to my issue / pull rquest and should be handled in a speparate issue, and the String concat in the log message I will not change.

Everything else is already done - I hope. :)

serious6:
there is the following:
> - Missing javadoc for param.
> - Is there a benefit for keeping whitespace for other BodyTypes? If not, then we might as well simplify the code and just never remove whitespace.
> @vboctor

superelchi:
- Missing javadoc in param: already done. See commit e00c76d.
- Never remove whitespace: this is something I can not decide. I'm not deep enought in your code to say yes or no. I'm using only the email send/receive part and have no knowledge (and can not test) if always keeping whitespace will break something in other parts. That is up to you gurus to decide. ;)

serious6:
maybe this should be held for further investigation. Meanwhile we can have this implementation with an optional param for handling the whitespaces merged I think. The question mentioned by @vboctor seems very hard to be confirmed while we dont have any tests that clarify this situation.

superelchi:
I agree.
So - everything resolved?

serious6:
@superelchi thanks for the contribution

serious6:
Created another issue for the handling of whitespaces in other BodyTypes.

PR #212, opened 2015-02-11
Handling the formatting issue

serious6:
I tried looking into the formatting-issue since we couldn´t just copy the google-style formatter cause of that license, I thought of letting contributors download the formatter and apply it to the project. Maybe anyone of you has a better idea on how to handle this. So assume this PR as a suggestion.
This PR also includes the following:
* last google_checks.xml that is working with the current checkstyle IntelliJ-plugin
* fixed checkstyle issues regarding the ... license header 
* reformatted the project sources


azurecla:

TTYL, AZPRBOT;

evpaassen:
We should really keep the codeStyleSettings in place. It has been **generated** by IntelliJ, right?

evpaassen:
And we should stop reformatting the source code over and over again. It makes the git history almost impossible to browse. Why is reformatting even required now? There's no explanation in this PR and the diff is far too large to browse the changes.

serious6:
codeStyleSettings is not generated its drawn together from the google-master. #158 declares there are  differences between (our settings.xml) intelliJ and eclipse formatter. So one goal should be to solve these issues and configure it right. 
As a assumption you prefer in doing these reformatting step by step (I would be fine with that too)?

vboctor:
Here are the priorities as far as I'm concerned:

- We get the big changes that will make merge impossible out of the way before 2.0 is out to avoid merge hell later between dev and stable branches.
- Have a decent coding style - doesn't have to be 100% consistent with google's.
- Have a way to stay clean in terms of style via integrations into the build and Travis.
- Style changes are in different checkins than functional commits.

Does that make sense?

As for the large diff, I would expect that local tools should be able to handle that rather than the github web interface.  So I don't see a reason to re-do the work and split this into multiple checkins, unless there are some other benefits.

In general though, once this is in, ideally, we are done with the formatting changes.

evpaassen:
What I tried to say was that *if* we reformat the entire thing *again*, we should make sure that we get the result right this time and that this will be the very last time reformatting the entire repo. The large changesets are really hard to handle when browsing history, not so much for the tools (ok, except you, GitHub), but more for humans trying to find when and why a specific change was made (blaming).

Reformatting has failed twice already, so this time we'll have to fix it once and for all.

evpaassen:
Furthermore, we really should keep the IntelliJ code style settings in the repo, or everyone will start submitting incorrectly formatted pull requests. If this file has not been generated by IntelliJ and causes legal clashes (I doubt it), we'll just have to create a new one using IntelliJ. And yes, the new file will most likely be 100% equal to what we have now.

serious6:
thanks for the feedback since we rely on the google-format.xml for intelliJ. I think we can't really guarantee that this file is free of bugs. I would also like having the formatter with our project-files but I remember @vboctor saying that this is prohibited by the license of that project. Is this information up to date?

serious6:
I will try to solve this with keeping the **codeStyleSettings**

PR #214, opened 2015-02-12
add sonatype parent pom for maven central releases

jangalinski:


azurecla:
Hi __@jangalinski__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

vboctor:
I'm not a maven expert.  It would be great if you can add a comment in the pom file and in the commit about what this solves.

serious6:
@jangalinski please provide a more suitable description for this PR.

serious6:
referencing the sonatype pom as parent will make the projects pom to inherit from its parents definition, which may be [here](http://repo.maven.apache.org/maven2/org/sonatype/oss/oss-parent/9/oss-parent-9.pom)

jangalinski:
Extending the sonatype parent is the easiest way to prepare publishing snapshots and releases to public maven repos (like central or sonatype snapshots). It is a good idea to include this in every java open source project. It wasn't there, so I added it. It wont be needed until you decide to publish artifacts (which I hope is soon :-))

serious6:
@jangalinski is this really a must-have or something like nice to have? Maybe you also have some experience if the repository section will also be inherited from parent-pom?

jangalinski:
If you want to publish to maven central, this is sort of "must have" ... although you could copy the relevant sections instead of inheriting them. If you do not plan to publish binaries, or publish them to a self hosted repository, you do not have to care.
And yes, the repository section is inherited ...

serious6:
@jangalinski thanks for the update so maybe this PR should be extended by removing the sections / dependencies which will be inherited and so far redundant. As far as I see lots of the used dependencies / plugins mentioned in oss-pom have newer versions now. Maybe these should be overridden. Or if keeping the older versions will also fit we should have this documented.

serious6:
@jangalinski are you planning to do any changes?

jangalinski:
Not really, no. You will get to this point again when you start thinking about publishing, its just not important enough to go on discussing/fixing/commenting. So I'll close.

Issue #223, opened 2015-02-24
NullPointerException while parsing response with EwsXmlReader


Nowheresly:
I'm facing NullPointerException when I try to read some responses received from an Exchange Server 2007 SP1.
Those NullPointerException happen while trying to read characters such as in the snippet below:
```java
Characters characters = (Characters) this.presentEvent;
//...
if (characters.getData().length() != 0) {   // NullPointerException thrown here
     elementValue.append(characters.getData());
}
```

In fact, nothing prevent the field data of a `javax.xml.stream.events.Character` class from being `null`. 
This class needs to be more robust to prevent such exception.

PR #224, opened 2015-02-24
#223 Avoid NullPointerException while parsing response

Nowheresly:
- check for null are added before using data field from javax.xml.stream.events.Character
see issue #223

### DIFF
--- a/src/test/java/microsoft/exchange/webservices/data/EwsXmlReaderTest.java
+++ b/src/test/java/microsoft/exchange/webservices/data/EwsXmlReaderTest.java
@@ -1,0 +1,100 @@
+/*
+ * The MIT License
+ * Copyright (c) 2012 Microsoft Corporation
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package microsoft.exchange.webservices.data;
+
+import org.junit.Assert;
+import org.junit.Before;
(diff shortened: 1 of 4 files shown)


azurecla:
    <span>
    </span>

TTYL, AZPRBOT;

Nowheresly:
I added the unit test as a separate commit since the unit test requires a new dependency, namely mockito.
This dependency has a test scope and thus is only required for test purposes.

src/test/java/microsoft/exchange/webservices/data/EwsXmlReaderTest.java
```diff
@@ -0,0 +1,77 @@
+package microsoft.exchange.webservices.data;
```
serious6: please add a license-header like shown [here](https://github.com/OfficeDev/ews-java-api/blob/master/src/main/java/microsoft/exchange/webservices/data/AbsoluteDateTransition.java). Plz remind the line between it and the package.

src/test/java/microsoft/exchange/webservices/data/EwsXmlReaderTest.java
```diff
@@ -0,0 +1,77 @@
...
+    impl = new EwsXmlReader(new ByteArrayInputStream(("<?xml version=\"1.0\" encoding=\"UTF-8\"?>"
+        + "<test></test>").getBytes("UTF-8")));
+    MockitoAnnotations.initMocks(this);
+
+  }
+
+  @Test
+  public void testReadValueWhenCharacterDataIsNull() throws Exception {
+
+    doReturn(false).when(presentEvent).isStartElement();
```
serious6: should we maybe split the mocking code from the underlying test?

serious6:
hi @Nowheresly thanks for you contribution, I think mockito makes sense. I have also seen some alternatives on the web. Maybe you can have some words about why mockito is the right one to choose?
Please also take into account that we should have some tests (in the future) which will enable us to make some unit-tests on other api-procedures.

src/test/java/microsoft/exchange/webservices/data/EwsXmlReaderTest.java
Nowheresly: Done. Sorry for the miss...

src/test/java/microsoft/exchange/webservices/data/EwsXmlReaderTest.java
Nowheresly: I usually write unit tests like this one:
- as instance variables, I declare all mocks needed for the test. The instance variable called 'impl' is *always* an instance of the class under test. All these instances variables *must* be refreshed before each test method execution.
- I declare a setUp method annotated with '@Before' which is in charge of refreshing all instances variables before each test and that contains as well all generic mock configuration (ie: all mock configuration that can be shared with all test methods)
- finally, I declare test methods where I follow the convention //given //when //then (see http://en.wikipedia.org/wiki/Behavior_Driven_Development).

In the case of this specific test, I expect all my mock declarations to be useful and needed to write other test methods. That's why it makes sens to me to keep this mocking code inside the test. 
But I'm obviously open to other suggestion!

Nowheresly:
Hi @serious6 
sure, there's a lot of alternatives to mockito: easymock, spock to name a few.
- spock (http://docs.spockframework.org/en/latest/#) is really really powerful. But since it's groovy based (a jvm scripting language) and the project currently does not used groovy at all...
- mockito is younger than its counterpart easymock. It is well designed and IMHO is simpler to use than easymock. You can see a comparison [here](https://code.google.com/p/mockito/wiki/MockitoVSEasyMock)

Finally, Mockito has some limitations (see the [FAQ](https://github.com/mockito/mockito/wiki/FAQ)) but has proven to be very useful.

serious6:
@Nowheresly maybe you can squash some of the commits. This would provide a better overview.

Nowheresly:
makes sense indeed !
you can find the new pull request here #231

serious6:
you can do a squash as simple as:
* git reset --soft HEAD~4
* git commit (with the files you want to commit as one or more commits)
* git push --force

PR #227, opened 2015-02-25
Fix keep-alive handling.

MikeN123:
We now try to keep the connection alive by simply consuming the full response. If that does not work, we revert to the old behaviour (to make sure we always close the underlying resources).

This is a bit of a hack, but the whole library is still build around the HttpClient 3 behaviour (where you had to close the request), whereas in HttpClient you have to consume the response (or close the request if there is no response, imho this is a flaw in the API of HttpClient 4).

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/HttpClientWebRequest.java
+++ b/src/main/java/microsoft/exchange/webservices/data/HttpClientWebRequest.java
@@ -18,10 +18,12 @@ import org.apache.http.auth.NTCredentials;
 import org.apache.http.client.CredentialsProvider;
 import org.apache.http.client.config.AuthSchemes;
 import org.apache.http.client.config.RequestConfig;
+import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpPost;
 import org.apache.http.client.protocol.HttpClientContext;
 import org.apache.http.config.SocketConfig;
 import org.apache.http.impl.client.*;
+import org.apache.http.util.EntityUtils;
 
 import java.io.*;
 import java.util.Arrays;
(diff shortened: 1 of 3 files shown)


azurecla:
Hi __@MikeN123__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!

TTYL, AZPRBOT;

src/main/java/microsoft/exchange/webservices/data/HttpClientWebRequest.java
```diff
@@ -57,11 +59,16 @@ public HttpClientWebRequest(CloseableHttpClient httpClient, HttpClientContext ht
...
+    // This releases the connection but keeps it alive for future requests
+    // If that is not possible, we simply cleanup the whole connection
+    if (response != null && response.getEntity() != null) {
+      EntityUtils.consume(response.getEntity());
+    } else if (httpPost != null) {
       httpPost.releaseConnection();
-      //postMethod.abort();
     }
+
     httpPost = null;
```
vboctor: Should this be inside the elseIf?  i.e. if you just consume the response should be leave the httpPost as is rather than setting it to null?

MikeN123: I don't think so. One of the invariables of this class seems to be something like:

`if (httpPost != null) then the response has not been closed yet`

So we have to set it to `null` to make sure we do not drop the connection if we call `close()` for a second time.
Maybe we also should set `response` to `null`, but I'm not entirely sure nothing in the library calls `getInputStream()` or `getErrorStream()` after `close()`, so I chose not to change that behaviour.

serious6: maybe this should be added as a comment to have this in mind for later changes?

MikeN123: I added a comment about this.

PR #228, opened 2015-03-01
Fixes #25 by replacing the custom buggy base64 implementation with Apache Common's

### LINKED ISSUE #25: Base64EncoderStream.encode() produces characters outside the Base64 alphabet
When invoking Base64EncoderStream.encode(), it will can produce a String containing characters outside of the Base64 alphabet. This should be fixed, calls replaced with Base64Encoder.encode(), or replaced entirely with a third-party library.

jasonparekh:
Like issue #25, I was running into a problem where the generated base64 was out-of-spec ('<' and ':' characters being produced.)  Instead of attempting to fix the custom logic, it's more expedient and safer to use an existing library.


azurecla:
Hi __@jasonparekh__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

serious6:
@jasonparekh please sign the contribution license agreement.

src/main/java/microsoft/exchange/webservices/data/Base64EncoderStream.java
```diff
@@ -23,69 +23,15 @@
...
 
+import org.apache.commons.codec.binary.Base64;
+import org.apache.commons.codec.binary.StringUtils;
+
 import java.util.Vector;
 
 /**
  * The Class Base64EncoderStream.
  */
 class Base64EncoderStream {
```
serious6: couldnt this class also been removed?

jasonparekh: It could have, was going for the least churn.  If you'd like, I'll inline the calls at all the call sites.

serious6: would be nice I think.

serious6:
hi @jasonparekh the provided commit-messages are not declaring the underlying changes in a way to easily know what has been changed. Please provide more sufficient commit-messages

azurecla:
__@jasonparekh__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

jasonparekh:
OK Andre, I will take a look at how the branch deviated, clean it up and
submit separate PRs

serious6:
@jasonparekh thanks

vboctor:
@jasonparekh this pull request seems to include several unrelated fixes.  Anyway to have them separated into different pull requests?

I agree that we should use standard implementations for encoding/decoding.  I'm surprised that this is not part of the Java framework.  Anyway, the tricky part with pulling dependencies is that the more packages we pull the more likely that we could be in a situation where the consuming app is using the same library but with different versions.  So I generally prefer to limit the # of such dependencies.  How easy would it be to just pickup the base 64 encoding class from the apache library?  (I'm assuming the library license is compatible).

I also don't think it is a bad idea to keep the class for encoding/decoding which would allow us to easily change the implementation in the future as needed.

These are just my thoughts, but I'm open to learn more about the thinking about this in the java world.

serious6:
@jasonparekh the unrelated commits mentioned in this PR are there because you upgraded your fork from your local master. Unfortunately this master was not up to date with the origin. If you update the fork to the latest version the unrelated commits will be gone.

PR #232, opened 2015-03-02
Optimize EwsXmlReader.readValue

vbauer:
I faced with a performance problem. To investigate this issue, I run `ExchangeService.findAppointments` many times and got the following summary statistic:

* ExchangeService.findAppointments - 102 sec.
    * EwsXmlReader.readValue - 21 sec.
        * String.format - 14 sec.

EwsXmlReader.readValue uses String.format always (even when it is completely unnecessary). This PR improves this situation and prevents all unnecessary calls of String.format in this method.

It is only the one place where we can improve `ExchangeService.findAppointments`. I decided to not send all improvements to simplify code review.

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/EwsXmlReader.java
+++ b/src/main/java/microsoft/exchange/webservices/data/EwsXmlReader.java
@@ -430,20 +430,19 @@ class EwsXmlReader {
    */
   public String readValue(boolean keepWhiteSpace) throws XMLStreamException,
       ServiceXmlDeserializationException {
-    String errMsg = String.format("Could not read value from %s.",
-        XmlNodeType.getString(this.presentEvent.getEventType()));
     if (this.presentEvent.isStartElement()) {
       // Go to next event and check for Characters event
       this.read(keepWhiteSpace);
       if (this.presentEvent.isCharacters()) {
-        StringBuffer elementValue = new StringBuffer();
+        final StringBuilder elementValue = new StringBuilder();
         do {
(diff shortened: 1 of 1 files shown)


azurecla:
    <span>
    </span>

TTYL, AZPRBOT;

azurecla:
__@vbauer__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

src/main/java/microsoft/exchange/webservices/data/EwsXmlReader.java
```diff
@@ -1121,4 +1121,17 @@ private static boolean isNullOrEmpty(String namespacePrefix) {
...
   }
 
+  /**
+   * Gets the error message which happened during {@link #readValue()}.
+   *
+   * @param details details message
+   * @return error message with details
+   */
+  private String getReadValueErrMsg(final String details) {
+    return String.format(
```
serious6: could't this also been used with simple string concatenation since this would also be faster?

vbauer: I am not sure that it will improve performance, because error message is not typical case (it should happen very rare in positive cases), but I can change it if you want.

vbauer:
I've just updated PR (it was merged with master).

vbauer:
Thank you for merge.

serious6:
thanks for the contribution ;)

PR #236, opened 2015-03-03
Convert DateTimeParser to DateTimeUtils

vbauer:
I continue my research with JProfiler (https://github.com/OfficeDev/ews-java-api/pull/232):

Class `DateTimeParser` creates DateTimeFormats & DateTimeFormatters (when it is instantiated). It is very expensive (see DateTimeFormat#forPattern & DateTimeFormatter#withZoneUTC). Class `DateTimeParser` is used in very critical places like `EwsServiceXmlReader`, so it is important to be as fast as possible.

Classes `DateTimeFormat` and `DateTimeFormatters` are thread-safe, so I converted DateTimeParser to utility class.

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/DateTimePropertyDefinition.java
+++ b/src/main/java/microsoft/exchange/webservices/data/DateTimePropertyDefinition.java
@@ -23,7 +23,7 @@
 
 package microsoft.exchange.webservices.data;
 
-import microsoft.exchange.webservices.data.util.DateTimeParser;
+import microsoft.exchange.webservices.data.util.DateTimeUtils;
 
 import java.util.Date;
 import java.util.EnumSet;
@@ -89,8 +89,7 @@ class DateTimePropertyDefinition extends PropertyDefinition {
   protected void loadPropertyValueFromXml(EwsServiceXmlReader reader, PropertyBag propertyBag)
       throws Exception {
     String value = reader.readElementValue(XmlNamespace.Types, getXmlElement());
-    propertyBag.setObjectFromPropertyDefinition(this, new DateTimeParser()
-        .convertDateTimeStringToDate(value));
+    propertyBag.setObjectFromPropertyDefinition(this, DateTimeUtils.convertDateTimeStringToDate(value));
   }
 
 
--- a/src/main/java/microsoft/exchange/webservices/data/EwsServiceXmlReader.java
+++ b/src/main/java/microsoft/exchange/webservices/data/EwsServiceXmlReader.java
@@ -23,11 +23,10 @@
 
 package microsoft.exchange.webservices.data;
 
-import microsoft.exchange.webservices.data.util.DateTimeParser;
+import microsoft.exchange.webservices.data.util.DateTimeUtils;
 
 import java.io.InputStream;
 import java.text.DateFormat;
-import java.text.ParseException;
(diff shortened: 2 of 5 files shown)


azurecla:

TTYL, AZPRBOT;

evpaassen:
I don't really like this approach. When I created the `DateTimeParser` (which is a more descriptive name, I think) with thread-safety in mind, so an instance can be reused. On many places, it is being reused, but there are two places where it isn't (`DateTimePropertyDefinition` and `UserConfigurationDictionary`). I rather think those are to be fixed.

vbauer:
> The Utility pattern is a software pattern that is used for utility classes that do not require instantiation and only have static methods. The stateless class is designated as static so that no instance can be created. Good candidates for utility classes are convenience methods that can be grouped together functionally.

This class does not have any mutable states except some immutable thread-safe fields and it should be immutable thread-safe util-class by design. I do not see any reason, why not to do it.

Also the current approach is a performance-killer (see test from #232): creation of  DateTimeFormat and DateTimeFormatters takes 23 seconds in the same test. It takes too much time under load testing.

`DateTimePropertyDefinition` and `UserConfigurationDictionary` are not the only classes where this class was used. Another problematic place in code is `EwsServiceXmlReader`.

MikeN123:
The fact that a utility pattern exists, does not make it a good pattern....

vbauer:
I'm completely agree, but the current code uses "workarounds" and still looks like util-class (DateTimePropertyDefinition:92):

```java
new DateTimeParser().convertDateTimeStringToDate(value)
```

serious6:
I don't know what would speak against this refactoring when it is faster. I also see no problem with the utility pattern here. If you think the old name is more descriptive maybe we can leave it but make that one a utility class.

evpaassen:
The Utility pattern is a bit of an anti-pattern. In this case, there's no need for it. There are two usages of the DateTimeParser class which need to be fixed in order to achieve higher performance.

vbauer:
I do not understand what benefits do you want to get using current solution?

Some thoughts about your possible optimization:

Lets imagine that you make DateTimeParser field for each class (DateTimePropertyDefinition, UserConfigurationDictionary, EwsServiceXmlReader), but it does not help, because DateTimeParser will be instantiated each time when this classes will be instantiated (and it will be very often).

I guess, that after this optimization, you will make DateTimeFormats & DateTimeFormatters fields as static fields (to prevent unnecessary instantiation) and we come back to util-class.

Ofc, you can use Object Pool pattern or something else to make the goal, but it is over-design. It will be easier to use KISS principle.

serious6:
@evpaassen when saying `the DateTimeParser class which need to be fixed in order to achieve higher performance` you mean there should be stored a local object of the formatter class instead of being created with every call?

serious6:
hi, I dont see a reason why the provided change would make use of the described antipattern. Since the aspect why this has been marked as a possible candidate is because there might be a refactoring of code which normally belongs to the surroundings of an instantiated object. In this situation this could be true if we  would have integrated the parsing process to a given Date-class which would handle its parsing (which also would be a good choice in my opinion). But with the change provided there has only been a refactoring of the parser itself in which I dont see a problem.

vbauer:
@serious6 Completely agree.

serious6:
@vbauer thanks for the contribution

PR #244, opened 2015-03-09
Move IFunctions in separate class

vbauer:
It allows to:
* remove some duplicated code
* re-use thread-safe instances of IFunction interface

I also added unit tests for this changes.

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/IFunctions.java
+++ b/src/main/java/microsoft/exchange/webservices/data/IFunctions.java
@@ -1,0 +1,103 @@
+/*
+ * The MIT License
+ * Copyright (c) 2012 Microsoft Corporation
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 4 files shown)


azurecla:

TTYL, AZPRBOT;

serious6:
@vbauer please also state what makes this important.

vbauer:
**1.** In the current approach we have a lot of unnecessary duplicated code. For example this function repeats 5-6 times and each time it takes 5 lines of code:
```java
new IFunction<Object,String>() {
  public String func(Object o) {
    return o.toString();
  }
}
```
It could be replaced by simple single line of code:
```java
IFunctions.ToString.INSTANCE
```

**2.** Anonymous functions contain unnecessary references to the outer-classes. New way clear this situation. We do not need these references - no need to use them.

**3.** It allows to prevent unnecessary mistakes, because we use only functions which are covered by unit tests.

**4.** Also it allows to prevent unnecessary creation of `IFunction` instances. It means that we spend zero time to instantiate it and we spend zero time to collect garbage.

serious6:
@vbauer thanks for the contribution

vbauer:
@serious6 Thanks for merge. :)

Issue #245, opened 2015-03-09
FileOutputStream is not closed in metod FileAttachment.load(String fileName)


kafa3227:
Saved file attachment can't delete after save, while FileOutputStream in metod FileAttachment.load(String fileName) is not closed.

```java  
public void load(String fileName) throws Exception {
    File fileStream = new File(fileName);
    FileOutputStream fos = new FileOutputStream(fileStream);
    this.loadToStream = fos;
    try {
      this.load();
    } finally {
      this.loadToStream.flush();
      <b>this.loadToStream.close();</b>
      this.loadToStream = null;
    }

    this.fileName = fileName;
    this.content = null;
    this.contentStream = null;
  }
```

PR #250, opened 2015-03-17
Fix issue #245: close stream after reading file content

Nowheresly:
Fix issue #245: close stream after reading file content
avoid useless downcast in ExchangeService.getAttachments method.

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/FileAttachment.java
+++ b/src/main/java/microsoft/exchange/webservices/data/FileAttachment.java
@@ -214,12 +214,17 @@ public final class FileAttachment extends Attachment {
    */
   public void load(String fileName) throws Exception {
     File fileStream = new File(fileName);
-    FileOutputStream fos = new FileOutputStream(fileStream);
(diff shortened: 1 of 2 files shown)


azurecla:

TTYL, AZPRBOT;

src/main/java/microsoft/exchange/webservices/data/FileAttachment.java
```diff
@@ -218,8 +218,13 @@ public void load(String fileName) throws Exception {
     this.loadToStream = fos;
```
serious6: looking at the code I saw 
```java
    FileOutputStream fos = new FileOutputStream(fileStream);
    this.loadToStream = fos;
```
fos is only used to be assigned but the following should also work?
```java
this.loadToStream = new FileOutputStream(fileStream);
```

Nowheresly: makes sense indeed
done !

serious6:
@Nowheresly thanks for your contribution

Issue #39, opened 2014-09-03
Package organization


evpaassen:
All 591 types in this library are located in the same package. This really doesn't help understanding how the library is organized.

serious6:
@vboctor , @evpaassen I am currently working on a refactoring of the project hierarchie and will create a PR within the next 3 days.

evpaassen:
OK. If possible, could you commit in small steps? That should make it easier to review.

vboctor:
@serious6 are there certain principles or goals that we are trying to achieve with the re-organization that are worth discussing here first?  Or do you want to tackle this in the pull request directly?

I would do the package re-organization change after finishing the PR for the code style and the followup checkin to fix all code based on the new style rules.

serious6:
I thought of 'making the api easier to understand' as the main goal. Having everything in the same package is like a mess in this case. Since we have many classes like on ews-managed-api I will take the structure from there and try a matching on our project.

vbauer:
:+1:

nickersd:
Possibly related to this is that I'm finding it not extensible.  I have the need to extend the various items to include our business uses, but the xml element name in the schema is not what the api expects.  Still looking into the best way to handle this....

Noblebrown:
@nickersd, do you mean adding your own properties to the API and extending objects? If so, I don't believe this can be done directly. You may have to create your own extended properties if you're trying to add additional properties/data and handle them as such if you want to extend existing classes.

nickersd:
@Noblebrown Yes I understand but  I've walked into an existing calendar implementation and don't have the luxury of designing from scratch so I'm looking for a more elegant way than the very hackish way that I've currently made it work.  I was thinking subclassing might  be the most elegant way but I could be wrong.

vboctor:
:+1: for small commits and mirroring structure from EWS Managed API.

serious6:
do you guys have any prefs on the organisation of the exceptions / enumerations packages? I just did a couple of attempts in having this refactored but enforced several problems which I am not sure on how to deal with:
* package autodiscover has multiple references in exceptions which will result in circular dependencies once the package will be torn appart on the other hand these classes also have several references in the other packages so how to deal with that?
* same with enumerations
* do you have any suggestions on splitting the exceptions / enumerations into other packages should it be like:
```txt
-\
  -core -\
             \- exception \-
                                  \- any core exceptions
            \- foo
            \- bar
  - dns -\
             \- exception \-
[...]

PR #253, opened 2015-03-21
fix the Project hierarchie issue #39

serious6:
I am currently adding commits here. While I am doing so you can see the result so maybe we can find problems earlier.


azurecla:

TTYL, AZPRBOT;

serious6:
*package attributes is ready to roll*

serious6:
*package autodiscover (subpackages follow) is ready to roll*

vboctor:
@serious6 I think we should have auto-discover a separate library all together.  The reason is that auto-discover can be shared between the Exchange Web Services (SOAP) and the Exchange REST APIs.  Just something to take into consideration.  We can of course do this in steps.

So application can just take a dependency on auto-discover to discover the end point then using REST, or do auto-discover and then use ews-java-api to connect to SOAP service.

serious6:
@vboctor ok are you fine when we will do this as an extra task or pr? The reason why I think this could be usefull is that this refactoring will be huge even without side effects.

serious6:
*package autodiscover completed*

vboctor:
@serious6 makes sense to have it as another step.  I just wanted to mention it to make sure that we take this into consideration as we are crafting the package.

serious6:
*package dns / credentials completed*

serious6:
*package enumerations completed*

serious6:
*package exceptions completed*

serious6:
*package properties completed*

serious6:
*package interfaces completed*

serious6:
*package search completed*

serious6:
*package sync completed*

serious6:
*package unifiedMessaging completed*

serious6:
*package notifications completed*

serious6:
*package misc completed*

serious6:
*package security completed*

serious6:
ok I think this is it. I will also move the tests into their desired folders tomorrow

serious6:
your comments on the change are welcome

serious6:
test classes have now also been moved

vboctor:
:+1: Thanks for doing this work.

I think it is unfortunate that we can have such organization while not exposing more internals of the implementation as public though.

In the C# world, I can organize code using namespaces and folder hierarchy, while only exposing the small number of methods / types that I need externally.  This also helps user to easily discover the classes that they are expected to use without being buried into a set of internally targeted classes / methods.

serious6:
Since we only have public, private, protected and package private here I think this is some kind of given by the project architecture. There might be a twist but we have to decide if we want stuff as to be just in one package or have it splitted for better understanding.

vbauer:
You did the really great work! I just want to add some thoughts about this changes (maybe it will be interesting for you):

1. Package is similar to a database table. A database relation should be named in the singular as its records are considered to be instances of the relation. I prefer to make name of packages in singular.
2. Another moment with packages: using of camel-case for naming (ex: configurationSettings, idConversion, timeZones). Maybe it will be better to use just a simple names (like "configuration" or "setting").
3. Package "autodiscover": maybe it will be better to move all error-classes, collection-classes into separate packages. 
4. Package "definition": it will be great to add package "primitive" for definitions like BoolPropertyDefinition, DoublePropertyDefinition, ByteArrayPropertyDefinition, etc..
5. We can simplify all package, if we add "base" sub-packages for abstract classes.
6. Maybe it will be useful to add some extra sub-packages for "interfaces" (ex: "delegate" for delegate-classes, "base" for IFunction, IPredicate, etc..)

serious6:
@vbauer thanks for the feedback. I think the mentioned point 1 may be depending on the personal style but I will be looking forward in having these things integrated. If you like feel free to add a PR on this one.

evpaassen:
I did not find time this review this yet, but I'd like to comment on the above.

> * Package is similar to a database table. A database relation should be named in the singular as its records are considered to be instances of the relation. I prefer to make name of packages in singular.

Here's the explanation: http://programmers.stackexchange.com/questions/75919/should-package-names-be-singular-or-plural

> * Another moment with packages: using of camel-case for naming (ex: configurationSettings, idConversion, timeZones). Maybe it will be better to use just a simple names (like "configuration" or "setting").

Camel-casing in package names is very uncommon indeed. Package names should be lower case, no dashes, no underscores.

> * We can simplify all packages, if we add "base" sub-packages for abstract classes.
> * Maybe it will be useful to add some extra sub-packages for "interfaces" (ex: "delegate" for delegate-classes, "base" for IFunction, IPredicate, etc..)

I don't see a real advantage of this. I'm also not aware of any other major Java project using such package organisation.

vbauer:
> Here's the explanation: http://programmers.stackexchange.com/questions/75919/should-package-names-be-singular-or-plural
Indeed.

> Camel-casing in package names is very uncommon indeed. Package names should be lower case, no dashes, no underscores.
Maybe I was unclear, but it was a point: to prevent using camel-case in package naming.

> I don't see a real advantage of this. I'm also not aware of any other major Java project using such package organisation.
Ex: joda-time. Yes, I agree that it is not very popular practice for existing Java projects, because most of them were designed a long time ago. They were very small/compact in the beginning of their life-cycle and after some time it was impossible to re-structure packages without breaking of backward compatibility. The idea of this way is the same: to minimize classes in packages using groups to prevent huge number of classes (and refactoring) in the future. Also it simplifies navigation, because classes are grouped by meaning (ofc). Maybe it is not necessary for now.

serious6:
Despite the mentioned stackoverflow post I looked around a little and bigger apis often use singular names. And I must say I feel much better with that. It might also be a not so simple question if all the files in the package are homogeneous or heterogeneous. And using singular names might also be more stable for later changes. The only thing I could not rename by now is interfaces since interface is a reserved name.

serious6:
@vboctor I also moved all autodiscover corresponding files to that package for later extraction

serious6:
@vbauer , @evpaassen maybe you can have a brief look at the package organisation? Maybe you also have a suggestion for dealing with the interfaces package (I am currently not that creative)?

serious6:
maybe we can also discuss if this should be merged with all the containing commits or if this should be squashed?

vbauer:
@serious6 I suggest to glue (squash) all commits. I'm not sure that it is important to show full history for developers.

Maybe it will be simpler to stop at the suggested organization of packages. If I understand correctly, the first official release will not be in the nearest future, so we can improve it later.

evpaassen:
> @vbauer , @evpaassen maybe you can have a brief look at the package organisation and also have a suggestion for dealing with the interfaces package (I am currently not that creative)?

I'll try to find some time soon.

> maybe we can also discuss if this should be merged with all the containing commits or if this should be squashed? Even though this might heavily increase the commit counts I would suggest in merge every commit for later documentation.

> @serious6 I suggest to glue (squash) all commits. I'm not sure that it is important to show full history for developers.

No, please don't squash. With one large commit, there's a high chance the rename tracking in git will be detecting renames incorrectly, due to certain similarity in some simple files. I don't see why a higher commit count would be a problem, as this **won't** make navigation through the revision tree harder when blaming.

vbauer:
> No, please don't squash. With one large commit, there's a high chance the rename tracking in git will be detecting renames incorrectly, due to certain similarity in some simple files. (Yes, I found this out the hard way...)

Never ever saw described situation. As I understand André did not rename files (only folders), so it should be fine. Anyway, it could be checked (squashing commits + folder comparison).

> I don't see why a higher commit count would be a problem, as this won't make navigation through the revision tree harder when blaming.

It is not for blaming, but to keep history clean. At the current moment there are 90 commits, so it will just complicate navigation.

serious6:
> Maybe it will be simpler to stop at the suggested organization of packages. If I understand correctly, the first official release will not be in the nearest future, so we can improve it later.

As inteded by [semantic versioning](http://semver.org) the next changes that will break compatibility will be in version `3.0`. If you guys dont mention any further changes and @vboctor also gives his go I think we can have this for now. 
In my opinion there are some goals that can be fixed later on:
* interfaces should not be placed in a seperated package [stack](http://stackoverflow.com/questions/1004980/should-interfaces-be-placed-in-a-separate-package)
* enumeration package should be more structured or this should also be moved to the main packages where they are used
* exception package should be more structured or this should also be moved to the main packages where they are used see package autodiscover for example

serious6:
@vbauer I think the main goal of having these commits integrated 'as is' would be to follow the changes that have been taken place. Once we will squash the single steps will be lost forever.

vboctor:
- I'm assuming for Java gurus this is an overall good change.  I mentioned my issues with it compared to C#.  More stuff becoming public.
- :+1: for not squashing changes
- :+1: for getting this in and we can iterate later, but ideally before v2 is out.
- :+1: for grouping interfaces/enumerations/exceptions where they logically belong vs. in a package on their own.

vbauer:
@serious6 Yes, agree. It makes sense, if it is really needed.

vbauer:
@serious6 Just wanted to say thank you again, great work!

serious6:
np. Also thanks for discussion.

nickersd:
Using the syntax that is in the documentation I suddenly started getting
the below.  I would post my code but it's a cut and paste from the readme.

microsoft.exchange.webservices.data.autodiscover.exception.AutodiscoverLocalException:
The Autodiscover service couldn't be located.
at
microsoft.exchange.webservices.data.autodiscover.AutodiscoverService.internalGetLegacyUserSettings(AutodiscoverService.java:743)
at
microsoft.exchange.webservices.data.autodiscover.AutodiscoverService.getLegacyUserSettings(AutodiscoverService.java:520)
at
microsoft.exchange.webservices.data.autodiscover.AutodiscoverService.internalGetLegacyUserSettings(AutodiscoverService.java:967)
at
microsoft.exchange.webservices.data.autodiscover.AutodiscoverService.getUserSettings(AutodiscoverService.java:1856)
at
microsoft.exchange.webservices.data.core.ExchangeService.getAutodiscoverUrl(ExchangeService.java:3672)
at
microsoft.exchange.webservices.data.core.ExchangeService.autodiscoverUrl(ExchangeService.java:3631)

evpaassen:
Could that be related to #270 (with PR #271)?

GouthamPatala:
I guess PR #271 might solve this. @nickersd, can you please try your code by pulling the code from https://github.com/GouthamPatala/ews-java-api, build the jar and replace the current ews jar you got with this newly built jar in the classpath.

PR #255, opened 2015-03-21
Remove unnecessary type casting

vbauer:
Class PropertyBag could be simplified:
```java
<T> T getObjectFromPropertyDefinition
```
instead of
```java
Object getObjectFromPropertyDefinition
```

It prevents a lot of unnecessary type casting / unnecessary code.

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/core/PropertyBag.java
+++ b/src/main/java/microsoft/exchange/webservices/data/core/PropertyBag.java
@@ -277,10 +277,10 @@ public class PropertyBag implements IComplexPropertyChanged, IComplexPropertyCha
    *                                 the property.
    * @return Property value. May be null.
    */
-  private Object getPropertyValueOrException(
+  private <T> T getPropertyValueOrException(
       PropertyDefinition propertyDefinition,
       OutParam<ServiceLocalException> serviceExceptionOutParam) {
-    OutParam<Object> propertyValueOutParam = new OutParam<Object>();
+    OutParam<T> propertyValueOutParam = new OutParam<T>();
     propertyValueOutParam.setParam(null);
     serviceExceptionOutParam.setParam(null);
 
@@ -312,11 +312,13 @@ public class PropertyBag implements IComplexPropertyChanged, IComplexPropertyCha
         // property
         ComplexPropertyDefinitionBase complexPropertyDefinition =
             (ComplexPropertyDefinitionBase) propertyDefinition;
-        Object propertyValue = complexPropertyDefinition
-            .createPropertyInstance(this.getOwner());
-        propertyValueOutParam.setParam(propertyValue);
+        ComplexProperty propertyValue = complexPropertyDefinition
+            .createPropertyInstance(getOwner());
+
+        // XXX: It could be dangerous to return complex value instead of <T>
+        propertyValueOutParam.setParam((T) propertyValue);
         if (propertyValue != null) {
-          this.initComplexProperty((ComplexProperty) propertyValue);
+          this.initComplexProperty(propertyValue);
           this.properties.put(propertyDefinition, propertyValue);
         }
       } else {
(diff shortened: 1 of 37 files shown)


azurecla:

TTYL, AZPRBOT;

serious6:
this pr breaks the 1.6 compatibility. We should discuss if this is worth it.

vbauer:
@serious6 Yes, I've already faced with it and fixed.

serious6:
@vbauer please also update this one

vbauer:
@serious6 Done.

src/main/java/microsoft/exchange/webservices/data/core/PropertyBag.java
```diff
@@ -758,7 +758,7 @@ public Object getObjectFromPropertyDefinition(PropertyDefinition propertyDefinit
         propertyDefinition, serviceExceptionOut);
     serviceException = serviceExceptionOut.getParam();
     if (serviceException == null) {
-      return propertyValue;
+      return (T) propertyValue;
```
serious6: The cast here is only necessary because the propertyValue has been assigned to Object? Maybe you can fix this aswell? I have also spotted other occurrences of object in this class.

serious6: Please also add unit-tests for PropertyBag

vbauer:
I've just added unit test, but it is not so useful as wanted. It's a little bit complicated to write clear unit test in this place..

PS: Sorry, I had to make a new commit to prevent erasing of history in this ticket.

vbauer:
Now I'm really not sure about this PR. From one hand, we clean a lot of source code, but we make possible problem with type casting in the PropertyBag class. From other hand, previous version of code could also fail on the type casting (outside the PropertyBag class).

Noblebrown:
The idea is sound, but this API needs support for generics/parameterized types from top to bottom. The PropertyDefinition hierarchy would need it for the PropertyBag autocasting to really work as you'd be passing PropertyDefinition<T> to ensure proper typing on both put and get.

vbauer:
@Noblebrown It was the initial idea to start working on this way, but I've faced some interesting moments in PropertyBag class (see XXX comment in the PR). `ComplexProperty` class spoils everything.

Noblebrown:
The API already assumes that any PropertyDefinition with a AutoInstantiateOnRead flag is a ComplexProperty and will throw an exception if it fails that assertion. I don't think there would be any added risk by casting to T from PropertyDefinition<T extends ComplexProperty> in that case, at least no more than there already was. It would already fail that assertion if it wasn't. In that case, T is already a ComplexProperty type and we can cast it to T without issue and return it.

vbauer:
@Noblebrown Yes, you've completely caught that I wanted to say! I just don't like this "incoherently" behavior in the core class.. It should be a little bit clearer/predictable in my opinion.

If there is no objections, maybe we can merge this PR and continue to work on API in scope of #272.

src/test/java/microsoft/exchange/webservices/data/core/PropertyBagTest.java
```diff
@@ -44,9 +46,21 @@
    */
   @Test(expected=ArgumentException.class)
   public void tryGetPropertyType() throws Exception{
+    PropertyBag pb = createPropertyBag();
+    pb.tryGetPropertyType(String.class, new RecurrencePropertyDefinition("test", "none", null, ExchangeVersion.Exchange2010_SP2), new OutParam<String>());
+  }
```
serious6: Maybe we should also add some positive tests for the changed methods. I was also wondering if the creation of propertybag could also be referenced `@before`?

serious6: Is there a chance you will also add maybe only some tests for PropertyDefinitions?

serious6:
@vbauer thanks for your contribution. We should definitely keep working at this like @Noblebrown mentions in #272

serious6:
and maybe there is also the possibility to add more tests on this topic.

PR #257, opened 2015-03-21
mvn autodeployment

serious6:
- added secure vars

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,11 @@ jdk:
 # enable container-based infrastructure
 sudo: false
 
+env:
+  global:
+    - secure: "V37/WY3//w4fQUxmtKXw9dsU4iH5Z6HvTpBkZj+5DaX5h2Lj98FzALvkGSt7KvqVKQBbgETjs3HTMQkz6hwbLgX2okLKCQbmhZu3vQLqkWe6KPqnO1NOKSaU4vhCoGpScNWMiVxrzCHLHZf5tQu7d+1WVMk5GNlgGqeKVSnz5LY="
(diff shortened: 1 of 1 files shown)


azurecla:

TTYL, AZPRBOT;

vboctor:
:+1:

serious6:
little bad that this can only be tested in production

vboctor:
It is for security reasons.  They don't want secure variables available to pull requests.  But I agree, would be nice to enable pull request processing when it is submitted by a repository owner for example.

vboctor:
I wonder if after a while, we will know which secure value is what.  Should we add a comment next to each one about its purpose?

serious6:
@vboctor maybe this will make attacks easier since for normal users this might not be a relevant information. This might also be the reason why travis also encrypts the secure-variables name rather then leaving it in plain text. For the issue dealing with gpg signed artifacts I made some local tests with my keybase secret which worked quite well. I therefore have send you an invitation to keybase since it´s looking quiet interesting in the order of having a central spot where users may validate files against.
So if you like give it a try. For my local test I followed these steps:
-  accept invitation and add basic information to [keybase.io](https://keybase.io) account
-  download / install keybase for cmd
  -  **windows:**
    -  download / install [gnupg](https://www.gnupg.org/download/)
    - (gnupg didnt create a temp folder on my system. maybe you will have to create it manually) -> *C:\ Users\<User>\.gnupg*
    -  download / install [node.js](https://nodejs.org/#download)
  -  **linux**
    -  `apt-get install gnupg`
    - `apt-get install nodejs-legacy npm`
  -  `npm install -g keybase-installer`
  -  `keybase-installer`
  -  `keybase version` 

nearly finished now import keys to the keyring:
- `keybase login`
- `gpg --list-keys` (you should now see the keybase key in your keyring)

finished. now test with maven:
* navigate to ews-java-api via cmd
* `mvn clean install -Dgpg.passphrase=MYKEYBASESECRETKEY`

PR #269, opened 2015-03-30
remove new ews-java-api.jar from repo

serious6:
it will otherwise always track uploading the cache

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,8 @@ env:
 cache:
   directories:
     - $HOME/.m2/repository
+before_cache:
+  - rm -rf $HOME/.m2/repository/com/microsoft/ews-java-api
 
 notifications:
   webhooks:


azurecla:

TTYL, AZPRBOT;

vboctor:
:+1: for the change.  Though the "track uploading the cache" side effect is not clear to me.  If you can clarify the commit comment, that would be great.

serious6:
ill update the commit message but only to explain what I meant:
every travis build runs `mvn install` this will have our jars assembled and then these jars will be deployed to the local travis repository (located at `$HOME/.m2/repository`) since with every mvn install these jars will change due to the changed api the complete repository will be uploaded to the cache. To prevent this we can delete the ews-java-api folder before the repository gets to cache (we will not need our own jar in the repo as well so the deletion has no side effects).

vbauer:
:+1:

serious6:
commit-msg is updated

PR #274, opened 2015-03-31
Add animal-sniffer-maven-plugin to check JVM compatibility

vbauer:


azurecla:

TTYL, AZPRBOT;

serious6:
Hi @vbauer can you provide a short introducion of that plugin and the resulting benefit?

vbauer:
@serious6 Hi, yes, of course. This plugin makes very simple, but useful thing - it checks that all used in project classes/methods are contained in the specified JVM's run-time library. maven-compiler-plugin can't do it (it just checks a source level).

It means that you can build project using Java 6, Java 7, Java 8, etc. and do not care that you can accidentally use some classes or methods that are available only in your JDK (for example: JDK 1.8). You can check compatibility with JDK 6 on your local machine (instead of checking it on Travis CI or installing JDK 1.6). Everything works in the auto-mode, so it will allow to minimize human-related errors.

pom.xml
```diff
@@ -235,6 +235,27 @@
...
             </plugin>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>animal-sniffer-maven-plugin</artifactId>
+                <version>1.14</version>
+                <configuration>
+                    <signature>
+                        <groupId>org.codehaus.mojo.signature</groupId>
+                        <artifactId>java16-sun</artifactId>
+                        <version>1.0</version>
```
serious6: Latest version might be [1.1](http://search.maven.org/#browse%7C-379312183). Or is there a reason you used version 1.0 here?

serious6:
Thanks this really makes sense. I was wondering if we maybe would like to have this with the travis build only since a failing build would lead to a discussion whether a restrictive test lifecycle wold maybe prevent contributors from using newer java functions from the beginning. so maybe there will be changes that are worth upgrading to a newer java version in the future.

vbauer:
Should I provide some additional information or something else? (ex: move plugin in separate profile)

serious6:
Hmm I am still wondering what benefit we will have with this. Since travis runs the build against all available jdks. The only thing I see is that we can skip building with multiple jdks. In this case we should answer the question which one will be more stable.

vbauer:
It is possible to:
- use new versions of JDK on your local machine (instead of installing JDK 6)
- check compatibility with JDK 6 on local machine instead of sending PR and checking it on Travis CI (with the following fixes and sending new PR, again and again..)

serious6:
@vbauer ok thanks for the contribution and for mentioning this plugin.

PR #275, opened 2015-03-31
Add slf4j as logging facade

### LINKED ISSUE #267: Define a logging strategy
In several classes, we can find statements such as:
System.out.println("something");
ex.printStackTrace();

These statements should be replaced with some logging statements such as
LOG.info("something");
LOG.error(ex);

Basically, in java we have 3 different logging facades:
- java.util.logging (jul)
- jakarta commons logging (jcl)
- simple logging for java (slf4j)

- jul has the advantage of being included in java core library. Unfortunately, most developers think this api is useless and not very flexible.
- jcl is, by far, one of the most popular logging library. It is already a dependency of ews-java-api! Its main drawback is its rather old API that does not leverage java 5+ features (such as var args).
- slf4j is now very popular as well. It provides several bridges so that it can be used with almost any other logging facades (such as jcl for example). It leverages java 5+. So instead of writing:
```
if(LOG.isDebugEnabled()) {
  LOG.debug("foo "+bar);
}
```
we can directly write:
```
LOG.debug("foo {}", bar);
```
In the first snippet, since the arguments are evaluated before executing the method, the runtime is generating a string that would never be used if the logging level is higher than debug. To avoid such overhead, developers have no choice but to add this ugly if statement.
In the second snippet, leveraging var args, no need to add an if statement, the string will be created only if the log level is debug.

I opened this issue so that we can decide if we want to use a logging framework and if yes, which one we want to use.
My reference goes to slf4j for its efficience & flexibility.
Any thoughts ?
(Of course I'm ready to do a PR for this issue)

Nowheresly:
commons-logging is now excluded
logback and jansi are added for test purpose only
all System.out.print are replaced with regular LOG.debug statements
all e.printStackTrace() are replaced with regular LOG.error statements
Fixes #267

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -83,6 +83,9 @@
         <junit.version>4.12</junit.version>
         <hamcrest-all.version>1.3</hamcrest-all.version>
         <mockito-core.version>1.10.19</mockito-core.version>
+        <slf4j.version>1.7.12</slf4j.version>
+        <logback.version>1.1.3</logback.version>
+        <jansi.version>1.11</jansi.version>
     </properties>
 
     <profiles>
@@ -164,6 +167,12 @@
             <groupId>org.apache.httpcomponents</groupId>
             <artifactId>httpclient</artifactId>
             <version>${httpclient.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>commons-logging</groupId>
+                    <artifactId>commons-logging</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         
         <dependency>
@@ -176,6 +185,13 @@
             <groupId>commons-logging</groupId>
             <artifactId>commons-logging</artifactId>
             <version>${commons-logging.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>${slf4j.version}</version>
         </dependency>
 
         <dependency>
@@ -204,6 +220,28 @@
             <version>${mockito-core.version}</version>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
+            <version>${logback.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>jcl-over-slf4j</artifactId>
+            <version>${slf4j.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.fusesource.jansi</groupId>
+            <artifactId>jansi</artifactId>
+            <version>${jansi.version}</version>
+            <scope>test</scope>
+        </dependency>
+
     </dependencies>
 
     <build>
--- a/.idea/libraries/Maven__ch_qos_logback_logback_classic_1_1_3.xml
+++ b/.idea/libraries/Maven__ch_qos_logback_logback_classic_1_1_3.xml
@@ -1,0 +1,13 @@
+<component name="libraryTable">
+  <library name="Maven: ch.qos.logback:logback-classic:1.1.3">
+    <CLASSES>
+      <root url="jar://$MAVEN_REPOSITORY$/ch/qos/logback/logback-classic/1.1.3/logback-classic-1.1.3.jar!/" />
+    </CLASSES>
+    <JAVADOC>
+      <root url="jar://$MAVEN_REPOSITORY$/ch/qos/logback/logback-classic/1.1.3/logback-classic-1.1.3-javadoc.jar!/" />
+    </JAVADOC>
+    <SOURCES>
+      <root url="jar://$MAVEN_REPOSITORY$/ch/qos/logback/logback-classic/1.1.3/logback-classic-1.1.3-sources.jar!/" />
+    </SOURCES>
+  </library>
+</component>
--- a/.idea/libraries/Maven__ch_qos_logback_logback_core_1_1_3.xml
+++ b/.idea/libraries/Maven__ch_qos_logback_logback_core_1_1_3.xml
@@ -1,0 +1,13 @@
+<component name="libraryTable">
+  <library name="Maven: ch.qos.logback:logback-core:1.1.3">
+    <CLASSES>
+      <root url="jar://$MAVEN_REPOSITORY$/ch/qos/logback/logback-core/1.1.3/logback-core-1.1.3.jar!/" />
+    </CLASSES>
+    <JAVADOC>
+      <root url="jar://$MAVEN_REPOSITORY$/ch/qos/logback/logback-core/1.1.3/logback-core-1.1.3-javadoc.jar!/" />
+    </JAVADOC>
+    <SOURCES>
+      <root url="jar://$MAVEN_REPOSITORY$/ch/qos/logback/logback-core/1.1.3/logback-core-1.1.3-sources.jar!/" />
+    </SOURCES>
+  </library>
+</component>
--- a/.idea/libraries/Maven__org_fusesource_jansi_jansi_1_11.xml
+++ b/.idea/libraries/Maven__org_fusesource_jansi_jansi_1_11.xml
@@ -1,0 +1,13 @@
+<component name="libraryTable">
+  <library name="Maven: org.fusesource.jansi:jansi:1.11">
+    <CLASSES>
+      <root url="jar://$MAVEN_REPOSITORY$/org/fusesource/jansi/jansi/1.11/jansi-1.11.jar!/" />
+    </CLASSES>
+    <JAVADOC>
+      <root url="jar://$MAVEN_REPOSITORY$/org/fusesource/jansi/jansi/1.11/jansi-1.11-javadoc.jar!/" />
+    </JAVADOC>
+    <SOURCES>
+      <root url="jar://$MAVEN_REPOSITORY$/org/fusesource/jansi/jansi/1.11/jansi-1.11-sources.jar!/" />
+    </SOURCES>
+  </library>
+</component>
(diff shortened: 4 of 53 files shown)


azurecla:

TTYL, AZPRBOT;

MikeN123:
I don't really agree on adding jcl-over-slf4j as a dependency - current users of the library will have to exclude it if they want to use JCL in any way, and it will affect all other application dependencies using JCL, so it introduces quite a mess in dependency management. If application developers add 1 other dependency that uses JCL (which are unfortunately quite common) and are not aware of this, it will create conflicts on the classpath.

Not adding jcl-over-slf4j, but adding slf4j, would result in two logging libraries being used, but at least we won't break other libraries/logging the user already has in place.

The other option would be to stick to JCL, which is not such a bad option imho, I'm not sure the advantages of slf4j outweigh the disadvantages.

evpaassen:
We should never ever add `jcl-over-slf4j` to the dependencies of this project.

Nowheresly:
Hi @MikeN123, @evpaassen 
thanks for taking time to review my PR.
> We should never ever add jcl-over-slf4j to the dependencies of this project.
...
> I don't really agree on adding jcl-over-slf4j as a dependency

Good point !
I removed the runtime dependency `jcl-over-slf4j`.
Instead, I updated my PR as follow:
- the scope of the dependency `jcl-over-slf4j` is now set to *test*, so we can see our log statements while running our unit tests but it has no impact on applications that use ews-java-api.
- add commons-logging as dependency with scope *compile*, so we can compile our project, but this dependency is no more a runtime dependency of ews-java-api. This way, the application that uses ews-java-api is in charge of defining its own logging strategy (either by adding `slf4j-jcl` or `jcl-over-slf4j`).

Looks better now ?

Nowheresly:
@MikeN123 
> I don't really agree on adding jcl-over-slf4j as a dependency - current users of the library will have to exclude it if they want to use JCL in any way, and it will affect all other application dependencies using JCL, so it introduces quite a mess in dependency management. If application developers add 1 other dependency that uses JCL (which are unfortunately quite common) and are not aware of this, it will create conflicts on the classpath.

I agree with your comment as I said previously. I updated my PR to follow your guideline.

> Not adding jcl-over-slf4j, but adding slf4j, would result in two logging libraries being used, but at least we won't break other libraries/logging the user already has in place.

Sure, I just added the compile scope to `commons-logging` to avoid exposing 2 logging facades as runtime dependency.

> The other option would be to stick to JCL, which is not such a bad option imho, I'm not sure the advantages of slf4j outweigh the disadvantages.

I'm not sure of what you mean about the disadvantages of slf4j. See the spring framework for example, they made the switch to slf4j during their last major version update. Since we are talking about the next major version update ews-java-api, I think it is probably the right time to consider doing the switch as well.

MikeN123:
What I mean with the disadvantages, is that we are now requiring two different logging libraries from our users (JCL and SLF4J), and I'm not sure that is justified by some varargs support in the logging library.

Spring never switched to SLF4J, and will probably stay on JCL forever. All JBoss stuff is even moving away from SLF4J and towards JBoss Logging.

You say:
> Sure, I just added the compile scope to commons-logging to avoid exposing 2 logging facades as runtime dependency.

I hope commons-logging is still a runtime dependency? (I'm not too sure about Maven, but isn't compile included in runtime scope by default?) Otherwise HttpClient will start throwing `ClassNotFoundException`'s if the user does not configure `jcl-over-slf4j` or `commons-logging`.

serious6:
Compile is the very default scope so when no scope is provided the dependency will implicitly be scoped with compile. Since this means that accessing those classes at compile time this will result in every user to include this dependency when it is at execution. Maybe the conclusion:
> (I'm not too sure about Maven, but isn't compile included in runtime scope by default?) Otherwise HttpClient will start throwing ClassNotFoundException's if the user does not configure jcl-over-slf4j or commons-logging.

Can be misunderstandable because in a maven related way the scope runtime means that this dependency is used at runtime but explicitly not at compile time.

Nowheresly:
@serious6 oh you're right! What I really meant was `provided`scope (the dependency is used at compile time but not at runtime). This way, the end user has to decide which dependency (either `commons-logging`or 'jcl-over-slf4j`) he wants to use.

@MikeN123 absolutely, the end-user has to add himself either `commons-logging`or `jcl-over-slf4j` as dependency, otherwise he will experience `ClassNotFoundException`. Nothing will come by default. Sure, this has to be documented...

Nowheresly:
Hi @MikeN123 
I realized that I may have proposed PR #275 too quickly, maybe giving you the impression that I wanted to stop discussion about the choice of the logging facade. Please let me say it wasn't my intention and to prove it, I just submitted another PR #278 where `commons-logging` is used as logging facade!

Nowheresly:
Now, back to our discussion:
> What I mean with the disadvantages, is that we are now requiring two different logging libraries from our users (JCL and SLF4J)

Well, not exactly. either you choose jcl or slf4j, but you cannot use both. If you choose slf4j, you will have to add jcl-over-slf4j, which is a dependency that almost all users of slf4j already have.

> Spring never switched to SLF4J, and will probably stay on JCL forever.

ok, maybe my terms were not enough accurate. Sorry for that...
Back in 2009, the spring team admitted they should have used slf4j instead of jcl. See their blog [here](http://spring.io/blog/2009/12/04/logging-dependencies-in-spring/)
Here are some excerpts from this blog post:

> If we could turn back the clock and start Spring now as a new project it would use a different logging dependency. Probably the first choice would be the Simple Logging Facade for Java (SLF4J), which is also used by a lot of other tools that people use with Spring inside their applications.

or even:

> SLF4J is a cleaner dependency and more efficient at runtime than commons-logging because it uses compile-time bindings instead of runtime discovery of the other logging frameworks it integrates. 
...
> Well it is, but it is optional, and it should behave better than the vanilla commons-logging with respect to class loader issues, notably if you are in a strict container like an OSGi platform. Allegedly there is also a performance benefit because the bindings are at compile-time not runtime.

What a praise for slf4j, right ?
Now let's go back in 2015. See their new projects [spring-data-jpa](https://github.com/spring-projects/spring-data-jpa/blob/master/src/main/java/org/springframework/data/jpa/repository/cdi/JpaRepositoryExtension.java), [spring-data-rest](ttps://github.com/spring-projects/spring-data-rest/blob/master/spring-data-rest-core/src/main/java/org/springframework/data/rest/core/config/RepositoryRestConfiguration.java), [spring-xd](https://github.com/spring-projects/spring-xd/blob/master/spring-xd-analytics/src/main/java/org/springframework/xd/analytics/metrics/redis/LoggingRecoveryCallback.java), they all use slf4j internally.

Admittedly, spring-boot uses jcl internally, but look what you get when you use the [spring-boot-starter-logging](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-starters/spring-boot-starter-logging/pom.xml). Yes, slf4j and jcl-over-slf4j are configured by default !

So it seems the spring team advice about slf4j is still relevant today.

> All JBoss stuff is even moving away from SLF4J and towards JBoss Logging.

Alright, I was not aware of that. But as far as I can see, JBoss decided not to use slf4j anymore because they had specific requirement not met by slf4j. See [here](http://docs.jboss.org/hibernate/orm/4.3/topical/html/logging/Logging.html) 

> Like SLF4J and Jakarta’s commons-logging, JBoss Logging is a "logging bridge" providing integration with numerous logging frameworks. JBoss Logging was chosen because of its i18n support and its support for "message ids". 

Just to sum up what I have in mind: I don't think choosing jcl as logging facade for ews-java-api is wrong. It's rather a conservative choice, yet perfectly valid.

Jcl has some drawbacks that we must keep in mind. The feedback of the spring team seems to me very valuable. If we choose jcl, it's a choice that we will have to assume probably forever.

MikeN123:
I don't really have a preference on SLF4J or JCL - JCL is a classloader mess, but we already have it as a dependency due to Apache HttpComponents, so we won't ever get rid of it.

Maybe someone else has an opinion on the SLF4J vs JCL debate?

I do think JCL should be a runtime and not a compile time dependency, otherwise it would also end up in the test dependencies, and we have both JCL and the SLF4J JCL bridge on the classpath during tests? Also, adding it as a runtime dependency makes sure we do not accidentally use it in the library (it is only necessary for HttpComponents).

serious6:
I think for the usability of ews-java-api we should keep this as easy as possible for people that wan't to use our api. Since we already have jcl as a dependency and regarding your conversation this might be the easiest one to be used. Maybe to sum up @Nowheresly thoughts on this (correct me if I am wrong) we should discuss if the mentioned **efficiency at runtime** will be worth switching to another library. To have a decision on this maybe we should have this compared in a small benchmark. 
As a sidenote what ever library we will use maybe the Information provided in this discussion is  worth to be added to the wiki in order to help others in applying a bridge to their logging-api or excluding our dependencies in their project. 
Sry for not helping you out with another opinion but I currently don't have that much of a preference or spotted something as _the argument_ for one of the suggestions.

Nowheresly:
> Since we already have jcl as a dependency and regarding your conversation this might be the easiest one to be used.

Correct.

> we should discuss if the mentioned efficiency at runtime will be worth switching to another library. To have a decision on this maybe we should have this compared in a small benchmark?

Well, if ews-java-api uses jcl internally and if the application is using jcl-over-slf4j to redirect all jcl log statements to slf4j, the overhead is negligible (we just add one level of indirection).

Since there's clear winner, maybe should we stick with jcl.
I updated my PR #278 to resolve all conflicts. Unless someone wants to add new elements to our discussion, feel free to merge it. We must go forward!

pom.xml
```diff
@@ -204,6 +218,28 @@
             <version>${mockito-core.version}</version>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
+            <version>1.1.3</version>
```
serious6: please also add logback and jansi to dependency attributes.

Nowheresly: done

serious6:
@Nowheresly thanks for your work here. So #278 is the candidate for being merged. If there are any more suggestions on this or the other PR feel free to reply. In the meantime I'll tag that PR for version 2.0.

serious6:
> tagged as Won't Fix since discussion points out to merge #278 / tag will be removed if other arguments are mentioned.

PR #278, opened 2015-04-02
Add jcl as logging facade

### LINKED ISSUE #267: Define a logging strategy

Nowheresly:
logback and jansi are added for test purpose only
all System.out.print are replaced with regular LOG.debug statements
all e.printStackTrace() are replaced with regular LOG.error statements
Fixes #267 
This is the same PR as #275 , but this time, commons-logging is used as logging facade.

Of course, either this one or #275 can be merged, but not both.

### DIFF
--- a/src/test/resources/logback-test.xml
+++ b/src/test/resources/logback-test.xml
@@ -1,0 +1,42 @@
+<configuration scan="true" debug="true">
+    <!--
+
+        The MIT License
+        Copyright (c) 2012 Microsoft Corporation
+
+        Permission is hereby granted, free of charge, to any person obtaining a copy
+        of this software and associated documentation files (the "Software"), to deal
+        in the Software without restriction, including without limitation the rights
+        to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+        copies of the Software, and to permit persons to whom the Software is
+        furnished to do so, subject to the following conditions:
+
+        The above copyright notice and this permission notice shall be included in
+        all copies or substantial portions of the Software.
+
+        THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+        IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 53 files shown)


azurecla:

TTYL, AZPRBOT;

serious6:
> tagged for 2.0 as mentioned in [comment](https://github.com/OfficeDev/ews-java-api/pull/275#issuecomment-89995721)

serious6:
@Nowheresly once upgraded, ready to merge.

Nowheresly:
done

src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,18 @@
+<configuration scan="true" debug="true">
```
serious6: can you also add project license-header here?

serious6:
@Nowheresly thanks. After you added the missing license-header we are ready :stuck_out_tongue_winking_eye:

serious6:
@Nowheresly thanks for your contribution

PR #282, opened 2015-04-04
Project hierarchie II - move interface classes

serious6:
second PR to fix the project hierarchie. This time interfaces are being moved to the places where other classes implement them. When multiple classes from different packages implement an interface I moved it to the parent package. 

_Ref.:_
* #253
* #39


azurecla:

TTYL, AZPRBOT;

src/main/java/microsoft/exchange/webservices/data/core/request/ResolveNamesRequest.java
```diff
@@ -36,7 +36,7 @@
 import microsoft.exchange.webservices.data.enumeration.ServiceErrorHandling;
 import microsoft.exchange.webservices.data.enumeration.XmlNamespace;
 import microsoft.exchange.webservices.data.exception.ServiceXmlSerializationException;
-import microsoft.exchange.webservices.data.interfaces.ILazyMember;
+import microsoft.exchange.webservices.data.core.ILazyMember;
```
vboctor: What makes IAsyncResult fall into misc while ILazyMember fall into core?

vboctor:
:+1:

src/main/java/microsoft/exchange/webservices/data/core/request/ResolveNamesRequest.java
serious6: hmm I did this because `IAsyncResult`has only 1 usage in implements clause and that one is located in the misc package and `ILazyMember` has no usage in implements clause but 5 usages in new anonymous inheritor creation which are located in core.
Since this was some kind of guessing which could be the best package are you suggesting to put them together to core/misc?

serious6:
@vboctor do you request a change on the `IAsyncResult` / `ILazyMember` decision?

vboctor:
Your approach makes sense for now.  I wonder of .NET concepts like ILazyMember and IAsyncResult have equivalent in Java.  Given the low number of usages, if there is, we should move to the native java approach to achieve these goals.

Issue #235, opened 2015-03-03
Remove Intellij IDEA specific files from project


vbauer:
I think it will be better to:
- remove ".idea" folder and ".iml"-file
- update .gitignore file

For Maven projects is completely unnecessary. Code style configuration (and other IDEA-specific files) could be placed in the some new folder like "misc" or "misc/IDEA".

serious6:
problem is with keeping the files the formatter will be used automatically since it is project default. If we'll have this seperated the developer has to set the formatter on it's own. 
This can be somehow dangerous since we don't have an automated checkstyle configuration (before this is possible we need all violations to be fixed) to see if the PR matches our formatting criterias. 
I have seen an intelliJ maven plugin but this one is outdated but maybe you have some other ideas?

vbauer:
Just to share my experience:

In all my projects I use [maven-checkstyle-plugin](http://maven.apache.org/plugins/maven-checkstyle-plugin) without any IDE formatters, so I can check my code and code of other people on CI server (instead of manual code review). Automatic analysis tools looks better in this case.

vbauer:
I want to add some other thoughts about this task. In the current situation files `.idea/codeStyleSettings.xml` and `.idea/misc.xml` are tracked by Git and Intellij Idea:

Idea IDE usually wins the war and changes this files in someway. After that Git finds the untracked changes and there is a problem: we have a dirty working branch (unstaged changes), so we have to do something like in the command-line:

* `git checkout .idea/codeStyleSettings.xml` and `git checkout .idea/misc.xml`
* or just `git checkout .idea`

As result, it is uncomfortable to work with Git branches in the terminal.

evpaassen:
Currently, we're following the [guidelines from the JetBrains docs](https://intellij-support.jetbrains.com/entries/23393067-How-to-manage-projects-under-Version-Control-Systems).

It's true that IntelliJ sometimes alters some of the files checked in, which is annoying. But I think  `.idea/codeStyleSettings.xml` is not one of them, unless you make changes to the code style settings yourself.

Using a tool (Maven plugin?) to generate a project config could be a solution. I know such a plugin exists for Gradle, but I don't know about Maven.

vbauer:
In my environment `.idea/codeStyleSettings.xml` and `.idea/misc.xml` are changed each time when I work with ews-java-api. I think it could happens, because we have different versions of Intellij IDEA, and IDE converts some parameters (or add some new defaults).

[...]

PR #283, opened 2015-04-04
Rewrite deprecated code in EwsSSLProtocolSocketFactory

vbauer:


azurecla:

TTYL, AZPRBOT;

serious6:
:+1: I am wondering if it might be useful to allow users of the api to pass in their own `HostnameVerifier`?

src/main/java/microsoft/exchange/webservices/data/core/EwsSSLProtocolSocketFactory.java
```diff
@@ -23,24 +23,26 @@
 
 package microsoft.exchange.webservices.data.core;
 
+import org.apache.http.conn.ssl.DefaultHostnameVerifier;
```
serious6: @vbauer unfortunately I missed updating `<httpclient.version>` to `4.4.1` because it is for some reasons not listed at [mvn repo](http://mvnrepository.com/artifact/org.apache.httpcomponents/httpclient). Maybe you can add this since it is thematically related (would save me creating another PR for this :trollface: )?

vbauer: @serious6 No problem.

vbauer:
Updated. I split changes in several commits for better understanding.

ews-java-api.iml
```diff
@@ -10,7 +10,7 @@
     </content>
     <orderEntry type="inheritedJdk" />
     <orderEntry type="sourceFolder" forTests="false" />
-    <orderEntry type="library" name="Maven: org.apache.httpcomponents:httpclient:4.4" level="project" />
+    <orderEntry type="library" name="Maven: org.apache.httpcomponents:httpclient:4.4.1" level="project" />
```
serious6: please also add/remove files under `.idea/libraries/`

src/main/java/microsoft/exchange/webservices/data/core/EwsSSLProtocolSocketFactory.java
serious6: thanks :+1:

serious6:
@vbauer do you also have any opinions on my previous [comment](https://github.com/OfficeDev/ews-java-api/pull/283#issuecomment-89817067)?

vbauer:
@serious6 It is already implemented here: https://github.com/vbauer/ews-java-api/commit/bb600d5984506bb8770cc99b1e6077561583525b

`EwsSSLProtocolSocketFactory` allows to configure `HostnameVerifier` and `ExchangeServiceBase` allows to override method `createConnectionSocketFactoryRegistry` to override socket factory.

ews-java-api.iml
vbauer: Thank you for the found mistake (https://github.com/OfficeDev/ews-java-api/issues/235), fixed.

serious6:
sry didn't have that on my mind. Maybe you can add this to the wiki with a small example? I think if possible there also could be added some javaDocs (ref. `EwsSSLProtocolSocketFactory`) :yum:

.gitignore
```diff
@@ -15,7 +15,9 @@
 # User-specific stuff:
 .idea/workspace.xml
 .idea/tasks.xml
+.idea/findbugs-idea.xml
```
serious6: :+1:

vbauer:
@serious6 No problem, I'll do it asap.

.gitignore
serious6: maybe also add `.idea/codeStyleSettings.xml`

vbauer: Added.

evpaassen: This won't work, because of https://github.com/OfficeDev/ews-java-api/issues/235#issuecomment-90041202.

ews-java-api.iml
evpaassen: I think the entire `.idea/libaries` can be removed and ignored, because IntelliJ generates it automatically from the Maven configuration.

.gitignore
vbauer: Definitely.

ews-java-api.iml
vbauer: I think the entire .idea folder can be removed and ignored, but let's move it out of the scope of this ticket.

.gitignore
serious6: Maybe we then should just commit the changed line with the groovy regarding style changes even though we won't need them. Just to not have this on the commit overview every time.

vbauer: Actually, IDE Talk, Findbugs and Javascript plugins also modify `iml` file..

vbauer:
Updated.

PS: I had to open `AutodiscoverService` and `ExchangeService` for extending. It could be also useful for other developers.

.gitignore
```diff
@@ -15,8 +15,12 @@
...
+.idea/codeStyleSettings.xml
 .idea/dictionaries
+.idea/shelf
 *.iws
+*.ipr
```
serious6: > .ipr file based format
> Share the project .ipr file and and all the .iml module files, don't share the .iws file as it stores user specific settings.
> [devnet.jetbrains](https://devnet.jetbrains.com/docs/DOC-1186)

vbauer: We do not use the old file-based format for project (ipr & iws), so both of them could be ignored.
https://www.jetbrains.com/idea/help/project.html

Directory-based format is used instead of it (iml & .idea).

serious6:
@vbauer thanks for the contribution. As mentioned we should discuss the idea folder issue later on.

PR #287, opened 2015-04-08
Fix project codestyle

serious6:
* reformat code with project formatter
* optimize imports
* minor javadoc / comment bugs fixed


azurecla:

TTYL, AZPRBOT;

vbauer:
I think it will be great to add Maven checkstyle plugin in scope of this ticket.

src/main/java/microsoft/exchange/webservices/data/attribute/Attachable.java
```diff
@@ -31,7 +31,6 @@
 /**
  * The Interface Attachable.
  */
-@Target(ElementType.TYPE)
-@Retention(RetentionPolicy.RUNTIME) public @interface Attachable {
+@Target(ElementType.TYPE) @Retention(RetentionPolicy.RUNTIME) public @interface Attachable {
```
evpaassen: Annotation style is incorrect (very unusual) everywhere. Should be one per line, above the block.
https://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s4.8.5-annotations

src/main/java/microsoft/exchange/webservices/data/core/service/ServiceObjectInfo.java
```diff
@@ -187,87 +175,76 @@ public Object createServiceObjectWithAttachmentParam(
...
-        new ICreateServiceObjectWithServiceParam() {
-          public Object createServiceObjectWithServiceParam(
-              ExchangeService srv) throws Exception {
-            return new MeetingRequest(srv);
-          }
-        }, new ICreateServiceObjectWithAttachmentParam() {
-          public Object createServiceObjectWithAttachmentParam(
-              ItemAttachment itemAttachment, boolean isNew)
+    this.addServiceObjectType(XmlElementNames.MeetingRequest, MeetingRequest.class,
+                              new ICreateServiceObjectWithServiceParam() {
```
evpaassen: I don't think we should align arguments, but I should check the Google guide for it.

evpaassen:
Is there a specific reason to reformat everything again? As I browse through the changes, it doesn't seem to be much better afterwards, honestly. Keep in mind that reformatting is not always perfect in IntelliJ, unfortunately. Reformatting multiple times can give different results, although it might converge.

Optimizing the imports and fixing the comments seems fine.

src/main/java/microsoft/exchange/webservices/data/attribute/Attachable.java
vbauer: I don't like some things in Google's code style guide, but in this point I agree with Erik. It will be better to put annotations one per line. It is commonly used practice.

serious6: so this seems to be another bug in the project formatter. :bug:

serious6:
@evpaassen the reason is to be able to have some validation like checkstyle with can verify the styleguide with every PR. But as we see this is just some sort of finding one error in the formatter after another. I will revert the changes that align the code and just keep the import optimization since this is really usefull.

vboctor:
I think we should pick the important rules and enforce them via tools like checkstyle.  Otherwise, we shouldn't get too caught up in getting this perfect.  At the end of the day, functional improvements deliver the most value to consumers of the library.

PR #290, opened 2015-04-12
Optimize EwsUtilities

vbauer:
Class `EwsUtilities` is used in many places. This PR suggests to optimize some moments in this class.

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/core/EwsUtilities.java
+++ b/src/main/java/microsoft/exchange/webservices/data/core/EwsUtilities.java
@@ -23,13 +23,12 @@
 
 package microsoft.exchange.webservices.data.core;
 
-import microsoft.exchange.webservices.data.core.service.item.Item;
-import microsoft.exchange.webservices.data.core.service.ServiceObject;
-import microsoft.exchange.webservices.data.core.service.ServiceObjectInfo;
-import microsoft.exchange.webservices.data.misc.TimeSpan;
 import microsoft.exchange.webservices.data.attribute.EwsEnum;
 import microsoft.exchange.webservices.data.attribute.RequiredServerVersion;
 import microsoft.exchange.webservices.data.core.request.HttpWebRequest;
+import microsoft.exchange.webservices.data.core.service.ServiceObject;
+import microsoft.exchange.webservices.data.core.service.ServiceObjectInfo;
+import microsoft.exchange.webservices.data.core.service.item.Item;
 import microsoft.exchange.webservices.data.enumeration.EventType;
 import microsoft.exchange.webservices.data.enumeration.ExchangeVersion;
 import microsoft.exchange.webservices.data.enumeration.FileAsMapping;
@@ -52,11 +51,15 @@ import microsoft.exchange.webservices.data.interfaces.ICreateServiceObjectWithSe
 import microsoft.exchange.webservices.data.interfaces.ILazyMember;
 import microsoft.exchange.webservices.data.interfaces.IPredicate;
 import microsoft.exchange.webservices.data.interfaces.ISelfValidate;
+import microsoft.exchange.webservices.data.misc.TimeSpan;
 import microsoft.exchange.webservices.data.property.complex.ItemAttachment;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
-import java.io.ByteArrayInputStream;
+import javax.xml.stream.XMLOutputFactory;
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.XMLStreamWriter;
+
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.lang.reflect.Field;
@@ -74,14 +77,10 @@ import java.util.TimeZone;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import javax.xml.stream.XMLOutputFactory;
-import javax.xml.stream.XMLStreamException;
-import javax.xml.stream.XMLStreamWriter;
-
 /**
  * EWS utilities.
  */
-public class EwsUtilities {
+public final class EwsUtilities {
 
   private static final Log LOG = LogFactory.getLog(EwsUtilities.class);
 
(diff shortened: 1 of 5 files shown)


azurecla:

TTYL, AZPRBOT;

src/main/java/microsoft/exchange/webservices/data/core/EwsUtilities.java
```diff
@@ -709,10 +646,9 @@ public static String boolToXSBool(Boolean value) {
    */
   public static <T extends Enum<?>> void parseEnumValueList(Class<T> c,
       List<T> list, String value, char... separators) {
-    EwsUtilities.EwsAssert(c.isEnum(), "EwsUtilities.ParseEnumValueList",
-        "T is not an enum type.");
+    EwsUtilities.EwsAssert(c.isEnum(), "EwsUtilities.ParseEnumValueList", "T is not an enum type.");
 
-    StringBuffer regexp = new StringBuffer("");
+    StringBuilder regexp = new StringBuilder("");
```
serious6: maybe initializing with a blank string is not the best option here.

src/main/java/microsoft/exchange/webservices/data/core/EwsUtilities.java
```diff
@@ -381,8 +332,7 @@ public static boolean stringEquals(String first, String second) {
    * @param message   The message to use if assertion fails.
    */
   public static void EwsAssert(boolean condition, String caller, String message) {
-    assert condition : String.format("[%s] %s",
-        caller, message);
+    assert condition : String.format("[%s] %s", caller, message);
```
serious6: I think we should get rid of basic java assert since this can be disabled and therefore should not be used in production code.

src/test/java/microsoft/exchange/webservices/data/core/EwsUtilitiesTest.java
```diff
@@ -35,14 +34,4 @@
   public void testGetBuildVersion() {
     Assert.assertEquals("Build version must be 0s", "0.0.0.0", EwsUtilities.getBuildVersion());
```
serious6: Retuning `0.0.0.0` as the current build-version might be wrong. Maybe we should read a property here which can be filled by travis with every build for instance. Or we should remove this entirely.

src/main/java/microsoft/exchange/webservices/data/core/EwsUtilities.java
```diff
@@ -519,25 +457,28 @@ public static Item createItemFromItemClass(
...
    *
    */
   public static Class<?> getItemTypeFromXmlElementName(String xmlElementName) {
-    return EwsUtilities.serviceObjectInfo.getMember().getXmlElementNameToServiceObjectClassMap()
-        .get(xmlElementName).getClass();
+    final ServiceObjectInfo member = EwsUtilities.SERVICE_OBJECT_INFO.getMember();
+    final Map<String, Class<?>> map =
+        member.getXmlElementNameToServiceObjectClassMap();
+
+    // XXX: It looks like a bug: itemClass.getClass() == Class.class
```
serious6: maybe a test should enlighten us :-) ?

src/main/java/microsoft/exchange/webservices/data/core/EwsUtilities.java
```diff
@@ -600,11 +545,11 @@ public static String formatLogMessage(String entryKind, String logEntry)
     XMLOutputFactory factory = XMLOutputFactory.newInstance();
     XMLStreamWriter writer = factory.createXMLStreamWriter(outStream);
     EwsUtilities.writeTraceStartElement(writer, entryKind, false);
-    writer.writeCharacters(System.getProperty("line.separator"));
+    writer.writeCharacters(SystemUtils.LINE_SEPARATOR);
```
serious6: > The line.separator System Property. Line separator ("\n" on UNIX).
Defaults to null if the runtime does not have security access to read this property or the property does not exist.
**This value is initialized when the class is loaded.** If System.setProperty(String,String) or System.setProperties(java.util.Properties) is called after this class is loaded, the value will be out of sync with that System property.

Maybe this can be a potential bug if the `LINE_SEPARATOR` is changed afterwards.

src/main/java/microsoft/exchange/webservices/data/core/EwsUtilities.java
```diff
@@ -767,60 +702,41 @@ public static String serializeEnum(Object value) {
...
+      Object o = Boolean.parseBoolean(value);
       return (T) o;
-    } else if (cls.isInstance(new String())) {
+    } else if (cls.isAssignableFrom(String.class)) {
       return (T) value;
-    } else if (cls.isInstance(Double.valueOf(0.0))) {
-      Object o = null;
-      o = Double.parseDouble(value);
+    } else if (cls.isAssignableFrom(Double.class)) {
+      Object o = Double.parseDouble(value);
```
serious6: maybe get rid of Double -> Object -> Object -> T

src/main/java/microsoft/exchange/webservices/data/core/EwsUtilities.java
```diff
@@ -862,30 +778,26 @@ public static String serializeEnum(Object value) {
    */
   public static void validateParamCollection(EventType[] eventTypes,
       String paramName) throws Exception {
-
     validateParam(eventTypes, paramName);
-
     int count = 0;
 
     for (Object event : eventTypes) {
```
serious6: perhabs `Object` can be switched to `EventType`

src/main/java/microsoft/exchange/webservices/data/core/EwsUtilities.java
```diff
@@ -922,23 +828,15 @@ public static String dateTimeToXSDateTime(Date date) {
...
     // The TimeSpan structure does not have a Year or Month
     // property, therefore we wouldn't be able to return an xs:duration
     // string from a TimeSpan that included the nY or nM components.
-
-    return String.format("%sP%sDT%sH%sM%sS", offsetStr, Math.abs(timeOffset
-        .getDays()), Math.abs(timeOffset.getHours()), Math
-        .abs(timeOffset.getMinutes()), Math
-        .abs(timeOffset.getSeconds())
-        + "." + Math.abs(timeOffset.getMilliseconds()));
+    return String.format("%sP%sDT%sH%sM%sS", offsetStr, Math.abs(timeOffset.getDays()),
```
serious6: I would suggest concatenation of strings for faster handling

serious6:
@vbauer thanks for your contribution.

src/main/java/microsoft/exchange/webservices/data/core/EwsUtilities.java
```diff
@@ -412,31 +362,23 @@ public static String getNamespaceUri(XmlNamespace xmlNamespace) {
    * @return the namespace from uri
    */
   public static XmlNamespace getNamespaceFromUri(String namespaceUri) {
-    if (namespaceUri.equals(EwsErrorsNamespace)) {
+    if (EwsErrorsNamespace.equals(namespaceUri)) {
       return XmlNamespace.Errors;
-    }
-    if (namespaceUri.equals(EwsTypesNamespace)) {
+    } else if (EwsTypesNamespace.equals(namespaceUri)) {
```
vboctor: Why the else if, given that the blocks from previous if returns?

vbauer: It takes less lines of code with the same performance.

src/test/java/microsoft/exchange/webservices/data/core/EwsUtilitiesTest.java
vbauer: Completely agree, but I think it will be better to create separate ticket for this issue.

PS: We could also use some Maven plugin to do it.

src/main/java/microsoft/exchange/webservices/data/core/EwsUtilities.java
vbauer: Agree.

src/main/java/microsoft/exchange/webservices/data/core/EwsUtilities.java
vbauer: Agree.

src/main/java/microsoft/exchange/webservices/data/core/EwsUtilities.java
vbauer: :+1:

src/main/java/microsoft/exchange/webservices/data/core/EwsUtilities.java
vbauer: :+1:

src/main/java/microsoft/exchange/webservices/data/core/EwsUtilities.java
vbauer: I don't think that we can remove casting (because we can't do direct cast from primitive type to T), but I've just found a bug: `Number.class` and `Double.class` are processed in the wrong order.

vbauer:
Updated.

vbauer:
Updated.

serious6:
Thanks. Maybe you can sum up the optimizations in one of the commit messages? They currently provide very basic information :-)

src/main/java/microsoft/exchange/webservices/data/core/EwsUtilities.java
```diff
@@ -1212,11 +1113,10 @@ public static void validateParamAllowNull(Object param, String paramName)
...
    * @throws Exception the exception
    */
-  public static void validateParam(Object param, String paramName)
-      throws Exception {
-    boolean isValid = false;
+  public static void validateParam(Object param, String paramName) throws Exception {
+    boolean isValid;
 
-    if (param != null && param instanceof String) {
+    if (param instanceof String) {
```
serious6: I think we will have a problem here if the param is null which seems to be valid when looking at the compete block
```java
 if (param instanceof String) {
      String strParam = (String) param;
      isValid = !strParam.isEmpty();
    } else {
->    isValid = param != null;
    }
```

vbauer: Actually `param != null && param instanceof String` is equivalent to `param instanceof String`, because `null instanceof String == false`. I just simplified this statement.

vbauer:
Updated.

serious6:
@vbauer thanks for your contribution.

vbauer:
@serious6 Thank you for merge!

PR #293, opened 2015-04-16
Fix parse problem in EwsUtilities

tomravn:
This is a fix for parsing Strings into different types. The isAssignableFrom() usages had the classes the wrong way round. This is fine when the classes are the same, but not for eg Number and Integer. This error caused the ExchangeServerInfo parsing from a SOAP header to fail.


azurecla:
Hi __@tomravn__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

src/main/java/microsoft/exchange/webservices/data/core/EwsUtilities.java
```diff
@@ -729,16 +729,16 @@ public static String serializeEnum(Object value) {
         }
       }
       return null;
-    } else if (cls.isAssignableFrom(Double.class)) {
+    } else if (Double.class.isAssignableFrom(cls)) {
       return (T) ((Double) Double.parseDouble(value));
-    } else if (cls.isAssignableFrom(Number.class))  {
+    } else if (Number.class.isAssignableFrom(cls))  {
       return (T) ((Integer) Integer.parseInt(value));
```
serious6: casting `Number` to `Integer` will fail the following tests:
```java
  @Test
  public void testParseBigInt() throws IllegalAccessException, ParseException, InstantiationException {
    Assert.assertEquals(BigInteger.TEN, EwsUtilities.parse(BigInteger.class, BigInteger.TEN.toString()));
  }

  @Test
  public void testParseBigDec() throws IllegalAccessException, ParseException, InstantiationException {
    Assert.assertEquals(BigDecimal.TEN, EwsUtilities.parse(BigDecimal.class, BigDecimal.TEN.toString()));
  }
```
maybe there should be a more general implementation here.

src/test/java/microsoft/exchange/webservices/data/core/EwsUtilitiesTest.java
```diff
@@ -54,6 +56,11 @@ public void testGetBuildVersion() {
   }
 
   @Test
+  public void testParseInt() throws IllegalAccessException, ParseException, InstantiationException {
+    Assert.assertEquals(new Integer(14), EwsUtilities.parse(Integer.class, "14"));
+  }
```
serious6: maybe you can also add tests for: `Boolean`, `Double`, `Date`... (including `Number` types). I think we should also extend the bounds (min,max e.g.).

azurecla:
__@tomravn__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

src/test/java/microsoft/exchange/webservices/data/core/EwsUtilitiesTest.java
tomravn: I'm happy to add more tests for the other types.

As for BigInteger etc, I can't see the tests you've included. And my changes wouldn't break that as it previously returned an Integer. All I've changed is the order of the classes in the isAssignableFrom() check

serious6: thanks. I just thought about having this fixed if we are already at this point. The unit-test wasn´t there before I just created it to underline the defect. If you don't want to fix this here please create an issue for it.

serious6:
@tomravn are you planning any changes?

tomravn:
Sorry, I got pulled onto another project for a few days.

I've added the extra tests. 

Before that though I made the project OSGi ready since we're using it in an OSGi container and needed it to have the correct MANIFEST.MF entries. I'm a bit new to pull requests but can see now that I should have made the related changes on their own branch and setup a pull request from there instead! I can see from your history that someone previously made the project OSGi ready and it was merged in but later reverted although I couldn't see why. This bundle has been tested in Apache Karaf and works fine so maybe it's worth merging that as well?

srusso:
We've just fixed this as well. We also fixed all the other calls of isAssignableFrom() in that method, not only the one for Integer.class is broken. Could you fix those as well?

tomravn:
They were all changed in the first commit

srusso:
Great, looking forward to the merge then :)

serious6:
@tomravn thanks for the changes. plz note that we decided to not include osgi components to the main project as these should be handled on a different branch (e.q.). See #177 for further information and discussion. For that reason changing the dependencies should be handled with a distinct PR and may be included in a `2.X+` release. If you want to reopen discussion on that topic plz create or comment on the existing issues related to this topic.

serious6:
@tomravn since there seems to be a little bit of urgency on this topic (due to several issue reports) will you devide the pr as mentioned? I think this should be merged within the next days.

serious6:
created another PR during the lag of feedback given here. #305
Closing this for now.

tomravn:
Sorry for the lack of response. I'm not full time on our project that requires this.

I've moved our OSGi changes to a different branch and reverted them on the branch this PR was for. I know this is closed and there's now an alternative PR open but this is OK for merging if that's still possible

serious6:
hi @tomravn thanks for your reply. The alternative has already been merged. Once we are fine with branching we will also release an osgi-version of ews-java-api

PR #294, opened 2015-04-16
Project hierarchie III

serious6:
second PR to fix the project hierarchie. This time exceptions are being moved to the places where they where thrown. 

_Ref.:_
* #253
* #39 

This PR is based on:
* #282


azurecla:

TTYL, AZPRBOT;

serious6:
I didnt know where to put the following exceptions best because they are thrown from all over the project:
* ServiceValidationException
* ServiceVersionException

maybe you guys have an idea on this?

vboctor:
@serious6 I don't think we should purely base where exception is homed based on where it is thrown.  I think we should categorize exceptions in two buckets:

- General exceptions - e.g. ServiceValidationException, ArgumentNullException, serialization exceptions, timeout exceptions.  Even if these are not thrown from a package today, we may want to do that in version 2.x.  Such exceptions can be included in some "low dependency" package in the dependency tree.  For example, an exceptions package, misc, core, or something along these lines.

- Domain specific exception - e.g. auto-discover specific exception, calendar specific exception, etc.  These should be located close to where they are thrown.  Because it is unlikely that somewhere else outside auto-discover where we would want to throw an auto-discover exception.  So it is easy to guess the location and the location should be stable.

serious6:
ok so closing this for now

serious6:
@vboctor hmm looking into the things you said this should be really close to my approach.
I created a more general exception type under 
```xml
 \- misc
        \-exception
                \- ArgumentException
                \- FormatException
                \- ...
```
and some other places like:
* security -> exception for Xml related Exceptions (this is close to ews-managed-api)
* property -> exception for AttachmentException and TimezoneDefinitionException
* core -> exception for lots of the Service...Exceptions and the ones used in core (as mentioned close to where they are thrown)
* autodiscover already has a exception package with the ones which I expected there.

I tried to devide the exceptions as close located to where they are thrown thats why I splitted these exceptions under **core** and **property**. I dont know if getting really closer makes sense because this will result in multiple single exception packages.

serious6:
> Because it is unlikely that somewhere else outside auto-discover where we would want to throw an auto-discover exception. So it is easy to guess the location and the location should be stable.

Unfortunately this is not the current state provided for example if you look at `ServiceRequestException` this has 3 new instance creation references in autodiscover and 12 in core.request. So guessing is not always easy. From my point of view regarding this particular situation we should on the one hand remove the occurences in one of the package or provide a new exception under autodiscover because once we put this only under core we will create a circular dependency.
This will also be a real pain regarding the enumerations because there are even more references all over the place.

PR #321, opened 2015-05-15
Fix #284 to enable adding headers (e.g. OAuth)

### LINKED ISSUE #284: OAuth with EWS Java APIs
Hi ,

We found there is Limitation in EWS Java APIs: EWS Java APIs that we download from https://github.com/OfficeDev/ews-java-api only Support EWS features till Exchange2010_SP2, it does have any special classes to pass OAuth token. EWS Managed APIs for .net has the Microsoft.Exchange.WebServices.Data.OAuthCredentials class through which we can pass OAuth token but till now EWS Managed Java APIs has not provided that class.

So, we need to pass OAuth token as authorization header. 
Also, when I tried passing “Bearer” type OAuth token as authorization header, service.getHttpHeaders().put("Authorization", "Bearer <token>"); but it gave me error saying "The request failed. HTTP header 'Authorization=Bearer <token>' isn't permitted. Only HTTP headers with the 'X-' prefix are permitted." 
How this limitation was resolved?
Because EWS Java API code is open source, We made some changes in the code, so they are now allowing OAuth bearer tokens as authorization header. 
We simply commented out the validate() method code in ExchangeServiceBase.java class, which was doing check for HTTP headers to be with ‘X-’ prefix and was not allowing ‘bearer’ type token to pass.
Now, we can pass OAuth token as,
service.getHttpHeaders().put("Authorization", "Bearer <token>");
--------------------------------------------------------------------------------------------------------------
Could someone from you team make this changes or   you can suggest any other solution.

Thanks & Regards,
Showkath.

vboctor:
This check was removed from EWS Managed API since 2012.  Removing the check enables adding headers like OAuth.

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
+++ b/src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
@@ -73,12 +73,6 @@ import java.util.TimeZone;
  * Represents an abstract binding to an Exchange Service.
  */
 public abstract class ExchangeServiceBase implements Closeable {
-
-  /**
-   * Prefix for "extended" headers.
-   */
-  private static final String ExtendedHeaderPrefix = "X-";
-
   /**
    * The credential.
    */
@@ -484,12 +478,6 @@ public abstract class ExchangeServiceBase implements Closeable {
    * @throws microsoft.exchange.webservices.data.exception.ServiceLocalException the service local exception
    */
   public void validate() throws ServiceLocalException {
-    // E14:302056 -- Allow clients to add HTTP request headers with 'X-' prefix but no others.
-    for (Map.Entry<String, String> key : this.httpHeaders.entrySet()) {
-      if (!key.getKey().startsWith(ExtendedHeaderPrefix)) {
-        throw new ServiceValidationException(String.format("HTTP header '%s' isn't permitted. Only HTTP headers with the 'X-' prefix are permitted.", key));
-      }
-    }
   }
 
   /**


azurecla:
Hi __@vboctor__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

vboctor:
@serious6 @nshowkath Please review the pull request as per our discussion on #284

src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
```diff
@@ -484,12 +484,6 @@ protected void setCustomUserAgent(String userAgent) {
    * @throws microsoft.exchange.webservices.data.exception.ServiceLocalException the service local exception
    */
   public void validate() throws ServiceLocalException {
```
serious6: maybe the method can be declared as `abstract` since it does not contain any default implementation anymore.

src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
```diff
@@ -484,12 +484,6 @@ protected void setCustomUserAgent(String userAgent) {
...
-    // E14:302056 -- Allow clients to add HTTP request headers with 'X-' prefix but no others.
-    for (Map.Entry<String, String> key : this.httpHeaders.entrySet()) {
-      if (!key.getKey().startsWith(ExtendedHeaderPrefix)) {
```
serious6: `static final String ExtendedHeaderPrefix` should also be deleted since it is not used anywhere else

serious6:
:+1:

src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
vboctor: I think the fact that the method is empty doesn't mean that we should necessarily make it an abstract class.  I'll leave it as "default validation is no-op".

src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
vboctor: Fixed.

nshowkath:
Thanks for providing fix for this issue , I will verify and confirm the same.

nshowkath:
Apologies , I missed to update the status , this fix is working for us 
Thanks a  lot .

vboctor:
Great, thanks @nshowkath

PR #334, opened 2015-05-25
Correction of isAssignableFrom usage inside tryGetValue

andrejusc:


azurecla:
Hi __@andrejusc__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

serious6:
@andrejusc thanks for your contribution. Currently the travis build is failing because of `ExtendedPropertyCollectionTest`. After a small run I noticed that the test initialises a `new ExtendedPropertyDefinition();` which will by default set mapiType as a String. This does the trick why we dont get any exception as expected. Maybe you can edit this test and add some other ones to prove  the correctness of your function.

andrejusc:
Modified test case so it could pass again

serious6:
@andrejusc can you please add some other test cases? positiv tests etc. for instance?

andrejusc:
Added one positive test case now. Hope that suits at least current contribution.

serious6:
thanks. One test is not much but better than nothing.

PR #340, opened 2015-06-03
travis deployment of site to gh-pages branch

serious6:
script will now also deploy the generated site to gh-pages branch after deploying javadoc and sources to maven central. if one of the mentioned steps fail the script will exit.

Once we generated a gh-page this will enable people to see the actual generated site artifacts and javadoc without executing these steps locally.

### DIFF
--- a/deploy_snapshot.sh
+++ b/deploy_snapshot.sh
@@ -21,28 +21,80 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
-if [ "$TRAVIS_REPO_SLUG" != "OfficeDev/ews-java-api" ]; then 
-	echo "[DEPLOY] Skipping snapshot deployment for repo:'$TRAVIS_REPO_SLUG'."
+set -e # Any subsequent(*) commands which fail will cause the shell script to exit immediately
+
+if [ "$TRAVIS_REPO_SLUG" != "OfficeDev/ews-java-api" ]; then
+    echo "[DEPLOY] Skipping snapshot deployment for repo:'$TRAVIS_REPO_SLUG'."
 elif [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
-	echo "[DEPLOY] Skipping snapshot deployment for a pull request."
+    echo "[DEPLOY] Skipping snapshot deployment for a pull request."
 elif [ "$TRAVIS_BRANCH" != "master" ]; then
-	echo "[DEPLOY] Skipping snapshot deployment for branch:'$TRAVIS_BRANCH'."
+    echo "[DEPLOY] Skipping snapshot deployment for branch:'$TRAVIS_BRANCH'."
 elif [ "$TRAVIS_SECURE_ENV_VARS" == "false" ]; then
-	echo "[DEPLOY] Skipping snapshot deployment due to TRAVIS_SECURE_ENV_VARS is set to '$TRAVIS_SECURE_ENV_VARS'."
+    echo "[DEPLOY] Skipping snapshot deployment due to TRAVIS_SECURE_ENV_VARS is set to '$TRAVIS_SECURE_ENV_VARS'."
 elif [ "$TRAVIS_JDK_VERSION" != "oraclejdk7" ]; then
-	echo "[DEPLOY] Skipping snapshot deployment for jdk:'$TRAVIS_JDK_VERSION'."
-else 
-	echo "[DEPLOY] Deploying snapshot for commit:'$TRAVIS_COMMIT' @ build-id:'$TRAVIS_BUILD_ID'"	
-	# create settings.xml
-	echo "<settings><servers><server><id>ossrh-snapshot</id><username>${OSSRHUSER}</username><password>${OSSRHPASS}</password></server></servers></settings>" > $HOME/.m2/settings.xml
-	# deploy
-	if [ -z "${GPG_PASSPHRASE+xxx}" ]; then
-	    echo "[INFO] Deploying unsigned artifacts"
-            mvn clean deploy --settings="$HOME/.m2/settings.xml" -Dmaven.test.skip=true
-        else
-            echo "[INFO] Deploying signed artifacts"
-            mvn clean deploy --settings="$HOME/.m2/settings.xml" -Dmaven.test.skip=true -Dgpg.passphrase=$GPG_PASSPHRASE
-        fi
-	# clean up
-	rm -f $HOME/.m2/settings.xml
+    echo "[DEPLOY] Skipping snapshot deployment for jdk:'$TRAVIS_JDK_VERSION'."
+else
+    echo "[DEPLOY] Deploying snapshot for commit:'$TRAVIS_COMMIT' @ build-id:'$TRAVIS_BUILD_ID'"
+    # create settings.xml
+    echo "<settings><servers><server><id>ossrh-snapshot</id><username>${OSSRHUSER}</username><password>${OSSRHPASS}</password></server></servers></settings>" > $HOME/.m2/settings.xml
+    # deploy
+    if [ -z "${GPG_PASSPHRASE+xxx}" ]; then
+       echo "[INFO] Deploying unsigned artifacts"
+       mvn clean deploy --settings="$HOME/.m2/settings.xml" -Dmaven.test.skip=true
+    else
+       echo "[INFO] Deploying signed artifacts"
+       mvn clean deploy --settings="$HOME/.m2/settings.xml" -Dmaven.test.skip=true -Dgpg.passphrase=$GPG_PASSPHRASE
+    fi
+    # clean up
+    rm -f $HOME/.m2/settings.xml
+    echo "[DEPLOY] Done deploying snapshot"
+
+    if [ -z "${GITHUB_USER+xxx}" ]; then
+       echo "[GH-PAGES] Deployment of gh-pages skipped. No GITHUB_TOKEN set."
+    else
+       # download hub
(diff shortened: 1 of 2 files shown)


azurecla:

TTYL, AZPRBOT;

serious6:
gh-pages branch is missing and GITHUB_TOKEN is missing as a secure variable.
[howto](https://help.github.com/articles/creating-an-access-token-for-command-line-use/)

serious6:
I am currently also evaluating a scenario where travis creates a PR on the gh-pages branch rather than doing the commits directly. These afterwards need to be merged manually but we dont need to provide the access-token. What do you think about this?

serious6:
PR is updated for this scenario.

deploy_snapshot.sh
```diff
@@ -21,28 +21,80 @@
...
-	# clean up
-	rm -f $HOME/.m2/settings.xml
-fi
+    echo "[DEPLOY] Skipping snapshot deployment for jdk:'$TRAVIS_JDK_VERSION'."
+else
+    echo "[DEPLOY] Deploying snapshot for commit:'$TRAVIS_COMMIT' @ build-id:'$TRAVIS_BUILD_ID'"
+    # create settings.xml
+    echo "<settings><servers><server><id>ossrh-snapshot</id><username>${OSSRHUSER}</username><password>${OSSRHPASS}</password></server></servers></settings>" > $HOME/.m2/settings.xml
+    # deploy
+    if [ -z "${GPG_PASSPHRASE+xxx}" ]; then
```
vboctor: What is the +xxx?  Is this debugging code?

deploy_snapshot.sh
```diff
@@ -21,28 +21,80 @@
...
+       echo "[INFO] Deploying unsigned artifacts"
+       mvn clean deploy --settings="$HOME/.m2/settings.xml" -Dmaven.test.skip=true
+    else
+       echo "[INFO] Deploying signed artifacts"
+       mvn clean deploy --settings="$HOME/.m2/settings.xml" -Dmaven.test.skip=true -Dgpg.passphrase=$GPG_PASSPHRASE
+    fi
+    # clean up
+    rm -f $HOME/.m2/settings.xml
```
vboctor: What happens if the script fails when calling mvn and exists.  Are we doing this as an extra step but travis cleans up anyway?  Should we not let the script exist on failure between persisting and deleting the secret?

deploy_snapshot.sh
```diff
@@ -21,28 +21,80 @@
...
+       echo "[GH-PAGES] Deployment of gh-pages skipped. No GITHUB_TOKEN set."
+    else
+       # download hub
+       mkdir -m777 -v -p "$HOME/download" ; cd "$HOME/download" && { curl -L https://github.com/github/hub/releases/download/v2.2.1/hub-linux-amd64-2.2.1.tar.gz | tar xz ; cd -; }
+
+       # init travis-user
+       [[ -d "$HOME/.config" ]] || mkdir -m777 -v -p $HOME/.config
+       echo -e "github.com:\n- user: ${GITHUB_USERNAME}\n  oauth_token: ${GITHUB_TOKEN}\n  protocol: https\n" > $HOME/.config/hub
+
+       git config --global user.email "travis@travis-ci.org"
```
vboctor: No need to use --global for git user name/email.

deploy_snapshot.sh
```diff
@@ -21,28 +21,80 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
-if [ "$TRAVIS_REPO_SLUG" != "OfficeDev/ews-java-api" ]; then 
-	echo "[DEPLOY] Skipping snapshot deployment for repo:'$TRAVIS_REPO_SLUG'."
+set -e # Any subsequent(*) commands which fail will cause the shell script to exit immediately
+
+if [ "$TRAVIS_REPO_SLUG" != "OfficeDev/ews-java-api" ]; then
```
vboctor: What is the motivation to block this script for forks?  But default forks won't have travis hooked, and if they do, then we should probably deliver the same behavior.  Independent of whether we die here or not, we should use the environment variable for the rest of the script.

deploy_snapshot.sh
```diff
@@ -21,28 +21,80 @@
...
+       git push -fq --progress origin gh-pages
+
+       echo "[GH-PAGES] Applying local changes"
+       git rm -rf --ignore-unmatch "./docs/snapshots/$TRAVIS_BRANCH"
+       mkdir -m777 -v -p "./docs/snapshots/$TRAVIS_BRANCH"
+
+       # copy all the builded stuff into the snapshot dir
+       cp -Rf "$HOME/OfficeDev/ews-java-api/target/site/*" "./docs/snapshots/$TRAVIS_BRANCH"
+
+       git add .
```
vboctor: Should we support adding/deleting files?  i.e.
    git add -A .

deploy_snapshot.sh
```diff
@@ -21,28 +21,80 @@
...
+       echo "[GH-PAGES] Committing changes to local repository"
+       git commit --author "travis-ci <travis@travis-ci.org>" -m "[TRAVIS]Deploy mvn site to [gh-pages/$TRAVIS_BRANCH] @ $TRAVIS_BUILD_NUMBER"
+       echo "[GH-PAGES] Pushing changes to origin gh-pages"
+       git push -fq --progress origin gh-pages
+
+       echo "[GH-PAGES] Creating a PR"
```
vboctor: What happens if there are no changes in git.  Is this going to make commit command fails and hence script exists?  If so, let's clarify in the comments.  The goal is no pull request if no change of course.

deploy_snapshot.sh
```diff
@@ -21,28 +21,80 @@
...
+       eval "$HOME/download/hub-linux-amd64-2.2.1/hub pull-request -m "$(printf "[TRAVIS] UPDATE [gh-pages/$TRAVIS_BRANCH] @ $TRAVIS_BUILD_NUMBER\n\n$TRAVIS_BUILD_NUMBER\nAutomated Pull Request :shipit:")" -b 'OfficeDev/ews-java-api:gh-pages' -h ${GITHUB_USERNAME}/ews-java-api:gh-pages"
+
+       # clean up
+       rm -f $HOME/.config/hub
```
vboctor: Do you want to delete the downloaded hub tool as well?

deploy_snapshot.sh
```diff
@@ -21,28 +21,80 @@
...
+
+       echo "[GH-PAGES] Finished site deployment. Pull request now needs to be merged by humans."
```
vboctor: If two pull requests created, the second will be inclusive of the first?  Does that mean we need to process them in a certain order to avoid issues?

vboctor:
Nice work.  I wonder if we are better off with just having a dedicated account that is used by Travis to do such checkins.  Even if this is compromised for some reason, we will get notifications and revert the commits.  Rather than having to always generate/process pull requests.  Thoughts?

deploy_snapshot.sh
serious6: hmm there is no real need in doing this. I just did this as an additional check for having ews-java-api deployed from an other travis build than the main one.

deploy_snapshot.sh
serious6: should be a copy & paste error. will be removed.

deploy_snapshot.sh
serious6: in general travis will clean up everything after the build has finished. Removing the files containing sensitive information was done only as an additional task if travis will "forget" to delete files after the build. I will create a trap for it.

serious6:
having the dedicated account should also work. In general the secure vars should be secured anyway so maybe I am just some kind of fearful xD But if these should get exposed once an attacker could do some hard damage and restoring everything from the reflog might work but can be a pain in the ass.

deploy_snapshot.sh
serious6: hub is only needed if we decide to create a PR out of the commit. Since it contains no sensitive data we may let travis clean up the dirt.

deploy_snapshot.sh
serious6: as ./docs/snapshots/$TRAVIS_BRANCH will be deleted with every PR this cant happen because at least `<!-- Generated by javadoc (1.8.0_45) on Wed Jun 10 19:22:17 CEST 2015 -->` on `api-docs-index.html` will be changed (if it is not generated at the same second). Until now I didnt test what will happen if this scenario fits but I expect that some git function will exit with return-code 1 and the script will exit.

deploy_snapshot.sh
serious6: yes. Older ones will have to be closed or merged before newer ones.

deploy_snapshot.sh
serious6: yes. we should use git add -A.

serious6:
integrated:
* +xxx removed
* added a trap for clean-up on exit
* removed --global
* use git add -A instead of git add .

not integrated:
* don't create a PR just do the commits with a dedicated user (as it is on discussion)
* motivation to block this script for forks (dont have pros or cons for allowing it)

serious6:
as for switching between PR and direct commit maybe we just should give the PR a try and see if this works or if its too annoying. Changing the process here will not affect the real application so this can be changed afterwards. I dont have any experiences here aswell.

vboctor:
I'm OK with trying this out.  :+1:

serious6:
@vboctor will you provide the needed params:
* creation of gh-pages branch (with maybe a dummy index.html)
* creation of a github-user for the use of the GITHUB_TOKEN
* provide the needed secure vars

serious6:
I created an empty gh-pages branch here [https://github.com/OfficeDev/ews-java-api/tree/gh-pages](https://github.com/OfficeDev/ews-java-api/tree/gh-pages) once merged travis will be able to add the files here.

serious6:
closed for now

PR #342, opened 2015-06-03
Minor bug fixes

vbauer:


azurecla:

TTYL, AZPRBOT;

serious6:
:+1:

src/main/java/microsoft/exchange/webservices/data/search/filter/SearchFilter.java
```diff
@@ -1085,17 +1085,6 @@ protected void internalValidate() throws ServiceValidationException {
       if (this.otherPropertyDefinition == null && this.value == null) {
         throw new ServiceValidationException(
             "Either the OtherPropertyDefinition or the Value property must be set.");
-      } else if (value != null) {
```
serious6: can you have some words on why you removed that block?

src/main/java/microsoft/exchange/webservices/data/property/definition/ComplexPropertyDefinitionBase.java
```diff
@@ -99,26 +99,15 @@ protected void internalLoadFromXml(EwsServiceXmlReader reader,
     boolean justCreated = getPropertyInstance(propertyBag, complexProperty);
     if (!justCreated && this.hasFlag(PropertyDefinitionFlags.UpdateCollectionItems,
         propertyBag.getOwner().getService().getRequestedServerVersion())) {
-      ComplexProperty c = (ComplexProperty) complexProperty.getParam();
-      if (complexProperty.getParam() instanceof ComplexProperty) {
-        c.updateFromXml(reader, reader.getLocalName());
+      Object c = complexProperty.getParam();
```
serious6: shouldnt the `getParam()` Method return `ComplexProperty` rather than `Object`?

src/main/java/microsoft/exchange/webservices/data/search/filter/SearchFilter.java
vbauer: The next line was:
```
if (!((value instanceof Object) || (value instanceof ISearchStringProvider)))
```

Any object is instance of Object class, so this expressions is always "true", it means that full expressions is always "false".

vbauer:
JFYI: I've just removed unnecessary "instanceof" operator in the last commit.

src/main/java/microsoft/exchange/webservices/data/search/filter/SearchFilter.java
serious6: so the `ISearchStringProvider` check was a bug?

vbauer: Yep... the first part of this expression is always "true".

serious6:
@vbauer thanks for your contribution

Issue #154, opened 2014-12-30
Migrating 'Getting Started with the EWS JAVA API'


serious6:
Hi there,
I was wondering if it might be usefull to migrate the [current guide](https://github.com/OfficeDev/ews-java-api/blob/master/readme.md) to the [projects wiki](https://github.com/OfficeDev/ews-java-api/wiki)? Since it offers a better overview while there could be added more tutorials (and other wiki stuff).
What do you think about this?

vboctor:
I think we should do that.  I just opened #203 for the same and missed this one.

Q1nt:
It will be pretty useful. Especially more tutorials.

PR #369, opened 2015-07-08
Update README.md: only provide short usage info

evpaassen:
The entire Getting Started Guide should be on the wiki, or in a separate docs folder, I think. We should only provide the most relevant info in the README.


azurecla:

TTYL, AZPRBOT;

evpaassen:
Perhaps some of the removed content should still be migrated...

evpaassen:
Ref #154.

readme.md
```diff
@@ -1,1109 +1,61 @@
...
+```
+
+And finally, add the dependency to your project's `pom.xml`.
+```xml
+<project>
+  <dependencies>
+    <dependency>
+      <groupId>com.microsoft.ews-java-api</groupId>
+      <artifactId>ews-java-api</artifactId>
+      <version>2.0-SNAPSHOT</version>
```
serious6: maybe exchange this with `X.X-SNAPHOT` to prevent future changes (same for groovy section).

serious6:
:+1:

readme.md
evpaassen: However, if we'd keep the version number here, this would be the easiest way to determine the current project version for new users.

Many other projects have it this way, for example:
* [Spring Framework](http://projects.spring.io/spring-framework/)
* [HikariCP](https://github.com/brettwooldridge/HikariCP)

The downside is indeed that with every release, the version number has to be updated in the README too.

serious6: I also can live with that until we have a nice gh-page :wink:

serious6:
@evpaassen please add `[ci skip]` somewhere in the commit msg. We do not need to trigger a build for this change.

evpaassen:
Done.

serious6:
@evpaassen thanks for your contribution.

PR #375, opened 2015-07-19
Use gradle instead of maven

serious6:
This PR disables maven support and moves the project to be managed with gradle. This enables us to:
* delete all files in /.idea 
* let gradle generate a pom.xml
* extend build.gradle easily

After merging this PR the project files will be generated with gradle. Just open a shell or cmd and execute `gradlew idea` on the project root. This will generate all needed files and also integrate google-style and file-encoding.


azurecla:

TTYL, AZPRBOT;

gradle.properties
```diff
@@ -0,0 +1,2 @@
+ossrhUsername=dummyId
```
vboctor: Do we need this file to be in source control? Or should we make it gradle.properties.sample?  This is to avoid the real creds being checked in by mistake.

vboctor:
:+1: to get rid of /.idea files from source control!

gradle.properties
serious6: travis will Need this file during the build. If we want to skip the checkin we Need to create it with a new build in the .travis.yml or add it to gitignore if we want to avoid later checkins.

serious6:
_ref. #235_

serious6:
closed for now

PR #392, opened 2015-08-05
Fix testGetMicrosoftTimeZoneNameBad() test case

vboctor:


azurecla:
Hi __@vboctor__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

src/test/java/microsoft/exchange/webservices/data/util/TimeZoneUtilsTest.java
```diff
@@ -52,10 +52,11 @@ public void testGetMicrosoftTimeZoneNameBad() {
     } catch (final IllegalArgumentException ignored) {}
```
serious6: while looking at this I was wondering what we want to do here? `Assert.fail(TimeZoneUtils.getMicrosoftTimeZoneName(null));` I think this test should handle a `@Test(expected = IllegalArgumentException.class` for the null vlaue

serious6:
@vboctor Maybe this test should be splitted and annotated with `@Test(expected = ...)` what do you think?

serious6:
@vboctor is this PR still needed?

Issue #398, opened 2015-08-10
Updated Appointment via EWS shows UTC instead of local TZ


bhagyagdesai:
Hi  All
I have researched for this issue and am unable to find a solution.

Updated Appointment via EWS shows UTC instead of local TZ in attendee's calender event notification mail 

Here is what's wrong:
in when it showing time in UTC  instead of local timezone ie (UTC+05:30) Chennai, Kolkata, Mumbai, New Delhi . both timing has 05:30 difference so it create confusion,Both timing should be same.
![image](https://cloud.githubusercontent.com/assets/10958579/9169787/36182252-3f7b-11e5-81d2-b3bb93463ed9.png)

rajeshpatil1012:
@itbw please look into this issue

itbw:
@rajeshpatil1012 I think this is related to #357. 

* According to https://msdn.microsoft.com/en-US/library/office/Dn789029%28v=EXCHG.150%29.aspx, if StartTimeZoneDefinition is not provided, Exchange server interpretes StartTime as UTC. 

* I think the following things had happened:
 * Start TimeZone was set to "India Standard Time"
 * ews-java-api library fails to created proper TimeZoneDefinition
 * Appointment will be written anyway, but the time will be set as UTC

* I'm working on bug from #357, so when it will be fixed, this one should be fixed too.

rajeshpatil1012:
@itbw Thanks for your reply

rajeshpatil1012:
Hi @itbw  i tried following code in ServiceRequestBase.java class just giving hard coded value then i got correct result
when try whih dynamic value setting bias i have issue, how to set Bias  in this java file 
I think that is only option to change api ServiceRequestBase.java file to set correct time in notification mail in java

// Emit the RequestServerVersion header
		if ((this.getService().getRequestedServerVersion().ordinal() == ExchangeVersion.Exchange2010_SP1.ordinal()))
        {
			

		writer.writeStartElement(XmlNamespace.Types,
				XmlElementNames.RequestServerVersion);
		writer.writeAttributeValue(XmlAttributeNames.Version, this
				.getRequestedServiceVersionString());
		writer.writeEndElement(); /// RequestServerVersion
		
		writer.writeStartElement(XmlNamespace.Types,XmlElementNames.TimeZoneContext);
	      
	      writer.writeStartElement(XmlNamespace.Types, XmlElementNames.TimeZoneDefinition);
	      
	      String timeZonePretty=null;
[...]

Issue #343, opened 2015-06-04
EwsUtilities.PATTERN_MONTH and PATTERN_MINUTES compiles same regex into pattern


ashwin244:
I was testing the EwsUtilities.java page, while calling the TimeSpan getXSDurationToTimeSpan(String xsDuration) function noticed - 

    
    // Month
    m = PATTERN_MONTH.matcher(xsDuration);
    LOG.debug(m.find());
    int month = 0;
    if (m.find()) {
      month = Integer.parseInt(m.group().substring(0,
          m.group().indexOf("M")));
    }

    // Minute
    m = PATTERN_MINUTES.matcher(xsDuration);
    LOG.debug(m.find());
    int minute = 0;
    if (m.find()) {
      minute = Integer.parseInt(m.group().substring(0,
          m.group().indexOf("M")));
    }

would generate the same value due to the same regular expression compiled into a pattern.
private static final Pattern PATTERN_MONTH = Pattern.compile("(\\d+)M");
private static final Pattern PATTERN_MINUTES = Pattern.compile("(\\d+)M");

Probably, one of them needs to be changed.

vbauer:
Just want to share some information for everybody:
This patterns were used before refactoring of the `EwsUtils` class (rev.: 3afa991bd67191ce3683c041df002332c3a301b4).

serious6:
will you provide a PR?

itbw:
It seems to be related to #357. All time zone periods with minutes in them are parsed in a wrong way. 
* For example TimeSpan for this one ```-PT3H30M``` will be parsed as:
```
negative true
year 0
month 30
day 0
hour 3
minute 30
seconds 0
```

* The big M for months and minutes seems to be correct according to ISO 8601 and the only difference is 'T' in the middle. For example this one ```P1Y2M10DT2H30M``` is a correct Time interval.

PR #406, opened 2015-08-17
EWSUtilities.getXSDurationToTimeSpan fix

### LINKED ISSUE #357: How to use setStartTimeZone method of Appointment class?
Hi All,
Can anyone share the code-snippets of using setStartTimeZone() method of Appointment class.

itbw:
* This pull request fixes the problem with wrong parsing of Bias in Period. 

* Fixes  #357, #398 and #343 

* Problem: 
 * TimeSpan for this one -PT3H30M will be parsed as:
```
negative true
year 0
month 30
day 0
hour 3
minute 30
seconds 0
```
* The big M for months and minutes seems to be correct according to ISO 8601 and the only difference is 'T' in the middle. For example this one P1Y2M10DT2H30M is a correct Time interval.

* Fix:
 * Two method are refactored getXSDurationToTimeSpan() and getXSDurationToTimeSpanValue().
 * For parsing of TimeSpan, an Joda-Time Period is now used instead of row Java Regex.
 * Period.parse() can automatically parse any value in form ```P1Y2M10DT2H30M```, except of leading '-' symbol. For parsing '-' Java Regex is still used. 

* Test
 * Corresponding tests are added to new class TimeSpanUtilsTest

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/core/EwsUtilities.java
+++ b/src/main/java/microsoft/exchange/webservices/data/core/EwsUtilities.java
@@ -50,8 +50,11 @@ import microsoft.exchange.webservices.data.core.exception.service.local.ServiceV
 import microsoft.exchange.webservices.data.core.exception.service.local.ServiceVersionException;
 import microsoft.exchange.webservices.data.misc.TimeSpan;
 import microsoft.exchange.webservices.data.property.complex.ItemAttachment;
+
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.joda.time.Period;
+import org.joda.time.format.ISOPeriodFormat;
 
 import javax.xml.stream.XMLOutputFactory;
 import javax.xml.stream.XMLStreamException;
@@ -230,14 +233,6 @@ public final class EwsUtilities {
   private static final String XML_SCHEMA_DATE_TIME_FORMAT = "yyyy-MM-dd'T'HH:mm:ss'Z'";
 
   private static final Pattern PATTERN_TIME_SPAN = Pattern.compile("-P");
-  private static final Pattern PATTERN_YEAR = Pattern.compile("(\\d+)Y");
-  private static final Pattern PATTERN_MONTH = Pattern.compile("(\\d+)M");
-  private static final Pattern PATTERN_DAY = Pattern.compile("(\\d+)D");
-  private static final Pattern PATTERN_HOUR = Pattern.compile("(\\d+)H");
-  private static final Pattern PATTERN_MINUTES = Pattern.compile("(\\d+)M");
-  private static final Pattern PATTERN_SECONDS = Pattern.compile("(\\d+)\\."); // Need to escape dot, otherwise it matches any char
-  private static final Pattern PATTERN_MILLISECONDS = Pattern.compile("(\\d+)S");
-
 
   private EwsUtilities() {
     throw new UnsupportedOperationException();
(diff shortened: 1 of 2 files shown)


azurecla:
Hi __@itbw__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!

TTYL, AZPRBOT;

src/main/java/microsoft/exchange/webservices/data/core/EwsUtilities.java
```diff
@@ -875,79 +870,50 @@ public static TimeSpan getXSDurationToTimeSpan(String xsDuration) {
...
+    debugMessage.append(" / ");
+    debugMessage.append("hour: " + hour);
+    debugMessage.append(" / ");
+    debugMessage.append("minute: " + minute);
+    debugMessage.append(" / ");
+    debugMessage.append("seconds: " + seconds);
+    debugMessage.append(" / ");
+    debugMessage.append("milliseconds: " + milliseconds);
+
+    LOG.debug("Parded TimeSpan: " + debugMessage.toString());
```
serious6: this seems to be a lot overhead for a debug message?!

src/main/java/microsoft/exchange/webservices/data/core/EwsUtilities.java
```diff
@@ -875,79 +870,50 @@ public static TimeSpan getXSDurationToTimeSpan(String xsDuration) {
...
+    long retval =
+        day * TimeSpan.DAYS + hour * TimeSpan.HOURS + minute * TimeSpan.MINUTES + seconds
+            * TimeSpan.SECONDS + milliseconds * TimeSpan.MILLISECONDS;
+    
     if (negative) {
       retval = -retval;
     }
-    return new TimeSpan(retval);
 
+    return new TimeSpan(retval);
```
serious6: as TimeSpan is initialized with milliseconds here you can also use:
`period.toStandardDuration().getMillis()` rather then computing everything on your own.

serious6: ref. [Period.html#toStandardDuration](http://joda-time.sourceforge.net/apidocs/org/joda/time/Period.html#toStandardDuration())

src/test/java/microsoft/exchange/webservices/data/util/TimeSpanUtilsTest.java
```diff
@@ -0,0 +1,65 @@
...
+  // Tests for EwsUtilities.getXSDurationToTimeSpan()
+
+  private static final String PERIOD_HOURS = "-PT13H";
+  private static final String PERIOD_HOURS_MINUTES = "-PT5H30M";
+  private static final String PERIOD_FULL = "PT2H30M59.0S";
+  private static final String PERIOD_FULL_NEGATIVE = "-PT2H30M59.0S";
+
+
+  @Test
+  public void testPeriodHours() {
```
serious6: maybe also add some negative tests (illegal ones), any tests for boundary values?

src/main/java/microsoft/exchange/webservices/data/core/EwsUtilities.java
itbw: @serious6 I had a closer at Joda Time Period. It seems, that it does not support duration with years and months, which makes sense, because year and months have different length. Here we have two methods: getXSDurationToTimeSpan(), which tries to include years and months, and getXSDurationToTimeSpanValue(), which is not. 
Is there any particular reason for having two methods or I can just merge leave one method getXSDurationToTimeSpanValue() and refactor usage of getXSDurationToTimeSpan() in other classes?

PR #407, opened 2015-08-17
Strip trailing . from DNS lookups when converting to autodiscover url

easel:
The root cause here is that some (all?) auto discovery DNS lookups will return a domain name with a trailing ".". Unfortunately I don't know the why's or wherefore's, but I've seen it in a debugger connecting to live office installs. This patch strips those trailing periods and allows the auto discover process to proceed normally.

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/autodiscover/AutodiscoverDnsClient.java
+++ b/src/main/java/microsoft/exchange/webservices/data/autodiscover/AutodiscoverDnsClient.java
@@ -24,10 +24,10 @@
 package microsoft.exchange.webservices.data.autodiscover;
 
 import microsoft.exchange.webservices.data.core.EwsUtilities;
-import microsoft.exchange.webservices.data.dns.DnsClient;
-import microsoft.exchange.webservices.data.dns.DnsSrvRecord;
 import microsoft.exchange.webservices.data.core.enumeration.misc.TraceFlags;
 import microsoft.exchange.webservices.data.core.exception.dns.DnsException;
+import microsoft.exchange.webservices.data.dns.DnsClient;
+import microsoft.exchange.webservices.data.dns.DnsSrvRecord;
 
 import javax.xml.stream.XMLStreamException;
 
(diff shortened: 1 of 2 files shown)


azurecla:
Hi __@easel__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

src/main/java/microsoft/exchange/webservices/data/autodiscover/AutodiscoverDnsClient.java
```diff
@@ -91,12 +91,14 @@ protected String findAutodiscoverHostFromSrv(String domain)
           "No appropriate SRV record was found.");
       return null;
     } else {
+      String hostName = dnsSrvRecord.getNameTarget();
+      if (hostName.length() > 0 && hostName.charAt(hostName.length()-1)=='.') {
```
serious6: please use `StringUtils.endsWith(.., ..)` its doing the same but looking way clearer.

serious6:
:+1: maybe also add some test-cases?

easel:
Ok, switched to .endsWith, factored out the validity checks and added test cases for them.

src/test/java/microsoft/exchange/webservices/data/autodiscover/AutodiscoverDnsClientTest.java
```diff
@@ -0,0 +1,42 @@
+package microsoft.exchange.webservices.data.autodiscover;
```
serious6: license header is missing

src/main/java/microsoft/exchange/webservices/data/autodiscover/AutodiscoverDnsClient.java
```diff
@@ -71,11 +71,27 @@ protected AutodiscoverDnsClient(AutodiscoverService service) {
   }
 
   /**
+   * Extracts a valid autodiscover hostname, if any, from a dns srv response.
+   *
+   * @param dnsNameTarget The hostname response returned by DNS
+   * @return Autodiscover hostname (will be null if lookup failed).
```
serious6: dont think this is really doing a lookup?

serious6:
is there an existing issue declaring the problem? If not please create one (for documentation purpose).

easel:
I'm not aware of an open issue for this one. I'll open one and update the documentation per your note. Thanks!

easel:
Ok, think we may be all set here as well. Thanks!

serious6:
@easel thanks for your contribution

PR #408, opened 2015-08-17
Only attempt to set dns server if provided

easel:
This patch simply avoids throwing an NPE if the dnsServerAddress is not provided. This allows fallback to the system resolver.

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/dns/DnsClient.java
+++ b/src/main/java/microsoft/exchange/webservices/data/dns/DnsClient.java
@@ -43,6 +43,23 @@ import java.util.List;
 public class DnsClient {
 
   /**
+   * Set up the environment used to construct the DirContext.
+   *
+   * @param dnsServerAddress
+   * @return
+   */
+  static Hashtable<String, String> getEnv(String dnsServerAddress) {
+    // Set up environment for creating initial context
+    Hashtable<String, String> env = new Hashtable<String, String>();
+    env.put("java.naming.factory.initial",
+            "com.sun.jndi.dns.DnsContextFactory");
+    if(dnsServerAddress != null && !dnsServerAddress.isEmpty()) {
+      env.put("java.naming.provider.url", "dns://" + dnsServerAddress);
+    }
+    return env;
+  }
+
+  /**
    * Performs Dns query.
    *
    * @param <T>              the generic type
@@ -58,15 +75,8 @@ public class DnsClient {
 
     List<T> dnsRecordList = new ArrayList<T>();
     try {
-
-      // Set up environment for creating initial context
-      Hashtable<String, String> env = new Hashtable<String, String>();
-      env.put("java.naming.factory.initial",
-          "com.sun.jndi.dns.DnsContextFactory");
-      env.put("java.naming.provider.url", "dns://" + dnsServerAddress);
-
       // Create initial context
-      DirContext ictx = new InitialDirContext(env);
+      DirContext ictx = new InitialDirContext(getEnv(dnsServerAddress));
 
       // Retrieve SRV record context attribute for the specified domain
       Attributes contextAttributes = ictx.getAttributes(domain,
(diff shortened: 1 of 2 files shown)


azurecla:
Hi __@easel__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

src/main/java/microsoft/exchange/webservices/data/dns/DnsClient.java
```diff
@@ -63,7 +63,9 @@
       Hashtable<String, String> env = new Hashtable<String, String>();
       env.put("java.naming.factory.initial",
           "com.sun.jndi.dns.DnsContextFactory");
-      env.put("java.naming.provider.url", "dns://" + dnsServerAddress);
+      if(dnsServerAddress != null) {
```
vboctor: Would it make sense to also make sure that the string is not blank?

vboctor:
Thanks @easel for your contribution.  It would be great if you can provide an associated test case(s).

src/main/java/microsoft/exchange/webservices/data/dns/DnsClient.java
easel: Probably wouldn't hurt.

easel:
I'll see if I can figure out the appropriate place for the tests, I forward ported this from the dawn of time and haven't kept on the development of the test cases =p

src/main/java/microsoft/exchange/webservices/data/dns/DnsClient.java
```diff
@@ -43,6 +43,23 @@
...
+   *
+   * @param dnsServerAddress
+   * @return
+   */
+  static Hashtable<String, String> getEnv(String dnsServerAddress) {
+    // Set up environment for creating initial context
+    Hashtable<String, String> env = new Hashtable<String, String>();
+    env.put("java.naming.factory.initial",
+            "com.sun.jndi.dns.DnsContextFactory");
+    if(dnsServerAddress != null && !dnsServerAddress.equals("")) {
```
serious6: maybe use `dnsServerAddress.isEmpty()`

serious6:
:+1: also for adding a test that reproduces the problem

easel:
Makes sense. Switched to .isEmpty.

src/main/java/microsoft/exchange/webservices/data/dns/DnsClient.java
```diff
@@ -43,6 +43,23 @@
 public class DnsClient {
 
   /**
+   * Set up the environment used to construct the DirContext.
+   *
+   * @param dnsServerAddress
+   * @return
+   */
+  static Hashtable<String, String> getEnv(String dnsServerAddress) {
```
serious6: any special reason to make this package-private?

serious6:
once some tests have been added this is ready to merge.

src/main/java/microsoft/exchange/webservices/data/dns/DnsClient.java
easel: I need the mechanism to be testable, but it's not really external API so I didn't want to make it public. What do you suggest?

easel:
Hmm. Tests were there and have disappeared. Will investigate.

src/main/java/microsoft/exchange/webservices/data/dns/DnsClient.java
serious6: the method also should be reachable writing the tests in the correct test-package. In my opinion this one should be private/protected and maybe not static?

easel: Isn't that the point of package private? Marking the methods private fails to compile for me, but perhaps I'm missing something.

easel:
Ok tests are here now, complete with license header. Also created an issue and referenced it in the commit log. Hopefully good to go now!

serious6:
@easel thanks for your contribution.

PR #420, opened 2015-08-24
AbsoluteDate and Time by TimeChange fix

itbw:
Pull request to fix bug in parsing AbsoluteDate and Time in TimeChange class

Bug:
* The method tryReadElementFromXml(EwsServiceXmlReader reader) tries to read XML elements AbsoluteDate and Time using  
```
SimpleDateFormat sdfin = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss");
```

Error:
 * Then parsing of Exchange 2007 SP1 server response with MeetingTimeZone element causes following error:
```
microsoft.exchange.webservices.data.core.exception.service.remote.ServiceRequestException: The request failed. Unparseable date: "03:00:00"
```

Fix Description:
* According to TimeChangeType https://msdn.microsoft.com/en-us/library/office/aa566047%28v=exchg.150%29.aspx and http://www.w3schools.com/schema/schema_dtypes_date.asp:
 * Time should be of type xs:time. See https://msdn.microsoft.com/en-us/library/office/aa565027%28v=exchg.150%29.aspx and for examples http://books.xmlschemata.org/relaxng/ch19-77311.html). 
 * The correct value returned by server will be  ```<t:Time>02:00:00</t:Time>```
 * AbsoluteDate should be of type xs:date value see https://msdn.microsoft.com/en-us/library/office/aa581577%28v=exchg.150%29.aspx and for examples http://books.xmlschemata.org/relaxng/ch19-77041.html

* The parsing using SimpleDateFormat was replaced with DatatypeConverter (http://docs.oracle.com/javase/7/docs/api/javax/xml/bind/DatatypeConverter.html), which can parse all forms of xs:date and xs:time. 

* Because EwsUtilities,dateTimeToXSDate() returns date using UTC Formatter, time zone of parsed calendar is also set to UTC, to avoid platform depend behaviour (different default Time Zones on system)

* Corresponding tests are also added.

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/property/complex/TimeChange.java
+++ b/src/main/java/microsoft/exchange/webservices/data/property/complex/TimeChange.java
@@ -32,11 +32,15 @@ import microsoft.exchange.webservices.data.core.enumeration.misc.XmlNamespace;
 import microsoft.exchange.webservices.data.core.exception.service.local.ServiceXmlSerializationException;
 import microsoft.exchange.webservices.data.misc.Time;
 import microsoft.exchange.webservices.data.misc.TimeSpan;
+
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
-import java.text.SimpleDateFormat;
+import java.util.Calendar;
 import java.util.Date;
+import java.util.TimeZone;
+
+import javax.xml.bind.DatatypeConverter;
 
 /**
  * Represents a change of time for a time zone.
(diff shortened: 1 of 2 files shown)


azurecla:
Hi __@itbw__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!

TTYL, AZPRBOT;

serious6:
@itbw thanks for your contribution :+1:

Issue #371, opened 2015-07-08
ServiceRequestException: The request failed. 40


choonchernlim:
When invoking the following code in 2 running threads:-

    @Override
    public Set<ExchangeAppointmentBean> query(final String calendarEmail,
                                              final LocalDate startDate,
                                              final LocalDate endDate) {
        final ExchangeService service;
        final FindItemsResults<Appointment> findResults;
        final ExchangeCredentials credentials = new WebCredentials("user", "pwd");

        try {
            service = new ExchangeService(ExchangeVersion.Exchange2010_SP2);
            service.setCredentials(credentials);
            service.setUrl(new URI("https://hostname/ews/exchange.asmx"));

            final FolderId folderId = new FolderId(WellKnownFolderName.Calendar, new Mailbox(calendarEmail));
            final CalendarFolder calendarFolder = CalendarFolder.bind(service, folderId);
            final CalendarView calendarView = new CalendarView(startDate.toDate(), endDate.toDate());

            findResults = calendarFolder.findAppointments(calendarView);
        }
        catch (Exception e) {
            throw CommonUtils.createException(LOGGER, e, "Unexpected error occurred when querying Exchange Server");
        }
       
       ....

      service.close();
    }    

... I get this error from the try-catch block:-

	Caused by: microsoft.exchange.webservices.data.core.exception.service.remote.ServiceRequestException: The request failed. 40
		at microsoft.exchange.webservices.data.core.request.SimpleServiceRequestBase.internalExecute(SimpleServiceRequestBase.java:74)
		at microsoft.exchange.webservices.data.core.request.MultiResponseServiceRequest.execute(MultiResponseServiceRequest.java:158)
		at microsoft.exchange.webservices.data.core.ExchangeService.bindToFolder(ExchangeService.java:500)
		at microsoft.exchange.webservices.data.core.ExchangeService.bindToFolder(ExchangeService.java:519)
		at microsoft.exchange.webservices.data.core.service.folder.CalendarFolder.bind(CalendarFolder.java:60)
		at microsoft.exchange.webservices.data.core.service.folder.CalendarFolder.bind(CalendarFolder.java:75)
		at myproject.module.pto.service.impl.MicrosoftExchangeServiceImpl.query(MicrosoftExchangeServiceImpl.java:78)
		... 6 more
	Caused by: java.lang.ArrayIndexOutOfBoundsException: 40
[...]

Issue #423, opened 2015-08-24
Custom HTTPClient


itbw:
Currently ews-java-api creates HTTPClient internally. It would be convenient to set a custom HTTPClient. It will bring following benefits:
* Setting custom Retry Strategies
* Setting custom Interceptors 
* Setting custom Authentication Strategies.

In general, it should make usage of library more flexible.

prophe05:
+

gokhanmansur:
+1

PR #428, opened 2015-08-26
Make ExchangeService more or less thread safe

candrews:
HttpClientContext isn't thread safe, and isn't meant for reuse, so create a new one each time it's needed
Use a shared cookied store per ExchangeService instance, clearing when necessary
ExchangeService is now thread safe, as long as certain properties (meant to be only set once at initialization time) are not changed oftentime while ExchangeService is being used
Safely manage HttpClient instances so they're lazy initialized used double checked locking
Made fields that can be final final, which improve understability and encourages thread safe design

Resolves gh-371

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
+++ b/src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
@@ -31,6 +31,7 @@ import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.security.GeneralSecurityException;
+import java.security.SecureRandom;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.util.Date;
@@ -40,6 +41,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Random;
 import java.util.TimeZone;
+import java.util.concurrent.ConcurrentHashMap;
 
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamWriter;
@@ -92,7 +94,7 @@ public abstract class ExchangeServiceBase implements Closeable {
   /**
    * The binary secret.
    */
-  private static byte[] binarySecret;
+  private static byte[] binarySecret = generateSessionKey();
 
   /**
    * The timeout.
@@ -132,24 +134,26 @@ public abstract class ExchangeServiceBase implements Closeable {
   /**
    * The requested server version.
    */
-  private ExchangeVersion requestedServerVersion = ExchangeVersion.Exchange2010_SP2;
+  private final ExchangeVersion requestedServerVersion;
 
   /**
    * The server info.
    */
   private ExchangeServerInfo serverInfo;
 
-  private Map<String, String> httpHeaders = new HashMap<String, String>();
-
-  private Map<String, String> httpResponseHeaders = new HashMap<String, String>();
+  private Map<String, String> httpHeaders = new ConcurrentHashMap<String, String>();
 
   private WebProxy webProxy;
 
-  protected CloseableHttpClient httpClient;
+  private volatile CloseableHttpClient httpClient;
 
-  protected HttpClientContext httpContext;
+  private final Object httpClientLock = new Object();
 
-  protected CloseableHttpClient	httpPoolingClient;
+  private final CookieStore cookieStore = new BasicCookieStore();
+
+  private volatile CloseableHttpClient	httpPoolingClient;
+
+  private final Object httpPoolingClientLock = new Object();
   
   private int maximumPoolingConnections = 10;
 
@@ -161,7 +165,7 @@ public abstract class ExchangeServiceBase implements Closeable {
   /**
    * Default UserAgent.
    */
-  private static String defaultUserAgent = "ExchangeServicesClient/" + EwsUtilities.getBuildVersion();
+  private static final String defaultUserAgent = "ExchangeServicesClient/" + EwsUtilities.getBuildVersion();
 
   /**
    * Initializes a new instance.
@@ -170,13 +174,11 @@ public abstract class ExchangeServiceBase implements Closeable {
    * every other constructor.
    */
   protected ExchangeServiceBase() {
-    setUseDefaultCredentials(true);
-    initializeHttpClient();
-    initializeHttpContext();
+	this(ExchangeVersion.Exchange2010_SP2);
   }
 
   protected ExchangeServiceBase(ExchangeVersion requestedServerVersion) {
-    this();
+    setUseDefaultCredentials(true);
     this.requestedServerVersion = requestedServerVersion;
   }
 
@@ -194,30 +196,6 @@ public abstract class ExchangeServiceBase implements Closeable {
     this.httpHeaders = service.getHttpHeaders();
   }
 
-  private void initializeHttpClient() {
-    Registry<ConnectionSocketFactory> registry = createConnectionSocketFactoryRegistry();
-    HttpClientConnectionManager httpConnectionManager = new BasicHttpClientConnectionManager(registry);
-    AuthenticationStrategy authStrategy = new CookieProcessingTargetAuthenticationStrategy();
-
-    httpClient = HttpClients.custom()
-      .setConnectionManager(httpConnectionManager)
-      .setTargetAuthenticationStrategy(authStrategy)
-      .build();
-  }
-
-  private void initializeHttpPoolingClient() {
-    Registry<ConnectionSocketFactory> registry = createConnectionSocketFactoryRegistry();
-    PoolingHttpClientConnectionManager httpConnectionManager = new PoolingHttpClientConnectionManager(registry);
-    httpConnectionManager.setMaxTotal(maximumPoolingConnections);
-    httpConnectionManager.setDefaultMaxPerRoute(maximumPoolingConnections);
-    AuthenticationStrategy authStrategy = new CookieProcessingTargetAuthenticationStrategy();
-
-    httpPoolingClient = HttpClients.custom()
-        .setConnectionManager(httpConnectionManager)
-        .setTargetAuthenticationStrategy(authStrategy)
-        .build();
-  }
-
   /**
    * Sets the maximum number of connections for the pooling connection manager which is used for
    * subscriptions.
@@ -230,6 +208,9 @@ public abstract class ExchangeServiceBase implements Closeable {
   public void setMaximumPoolingConnections(int maximumPoolingConnections) {
     if (maximumPoolingConnections < 1)
       throw new IllegalArgumentException("maximumPoolingConnections must be 1 or greater");
+    if(httpPoolingClient!=null){
+    	throw new IllegalStateException("Cannot change the maximumPoolingConnections setting after a request has been made");
+    }
     this.maximumPoolingConnections = maximumPoolingConnections;
   }
 
@@ -252,29 +233,37 @@ public abstract class ExchangeServiceBase implements Closeable {
     }
   }
 
-  /**
-   * (Re)initializes the HttpContext object. This removes any existing state (mainly cookies). Use an own
-   * cookie store, instead of the httpClient's global store, so cookies get reset on reinitialization
-   */
-  private void initializeHttpContext() {
-    CookieStore cookieStore = new BasicCookieStore();
-    httpContext = HttpClientContext.create();
+  protected HttpClientContext createHttpClientContext() {
+	HttpClientContext httpContext = HttpClientContext.create();
     httpContext.setCookieStore(cookieStore);
+    return httpContext;
   }
 
   @Override
   public void close() {
-    try {
-      httpClient.close();
-    } catch (IOException e) {
-      LOG.debug(e);
+    if(httpClient!=null) {
+	  synchronized(httpClientLock){
+	    if(httpClient!=null) {
+	      try {
+	        httpClient.close();
+	      } catch (IOException e) {
+	        LOG.debug(e);
+	      }
+	      httpClient = null;
+	    }
+      }
     }
 
     if (httpPoolingClient != null) {
-      try {
-        httpPoolingClient.close();
-      } catch (IOException e) {
-        LOG.debug(e);
+      synchronized(httpPoolingClientLock){
+    	if (httpPoolingClient != null) {
+          try {
+            httpPoolingClient.close();
+          } catch (IOException e) {
+            LOG.debug(e);
+          }
+          httpPoolingClient=null;
+    	}
       }
     }
   }
@@ -322,7 +311,8 @@ public abstract class ExchangeServiceBase implements Closeable {
       throw new ServiceLocalException(strErr);
     }
 
-    HttpClientWebRequest request = new HttpClientWebRequest(httpClient, httpContext);
+    HttpClientWebRequest request = new HttpClientWebRequest(getHttpClient(), createHttpClientContext());
+    request.setProxy(getWebProxy());
     prepareHttpWebRequestForUrl(url, acceptGzipEncoding, allowAutoRedirect, request);
 
     return request;
@@ -352,11 +342,8 @@ public abstract class ExchangeServiceBase implements Closeable {
       throw new ServiceLocalException(strErr);
     }
 
-    if (httpPoolingClient == null) {
-      initializeHttpPoolingClient();
-    }
-
-    HttpClientWebRequest request = new HttpClientWebRequest(httpPoolingClient, httpContext);
+    HttpClientWebRequest request = new HttpClientWebRequest(getHttpPoolingClient(), createHttpClientContext());
+    request.setProxy(getWebProxy());
     prepareHttpWebRequestForUrl(url, acceptGzipEncoding, allowAutoRedirect, request);
 
     return request;
@@ -383,8 +370,6 @@ public abstract class ExchangeServiceBase implements Closeable {
     prepareCredentials(request);
 
     request.prepareConnection();
-
-    httpResponseHeaders.clear();
   }
 
   protected void prepareCredentials(HttpWebRequest request) throws ServiceLocalException, URISyntaxException {
@@ -644,8 +629,8 @@ public abstract class ExchangeServiceBase implements Closeable {
     this.credentials = credentials;
     this.useDefaultCredentials = false;
 
-    // Reset the httpContext, to remove any existing authentication cookies from subsequent request
-    initializeHttpContext();
+    // Reset the cookies, to remove any existing authentication cookies from subsequent request
+    cookieStore.clear();
   }
 
   /**
@@ -673,8 +658,8 @@ public abstract class ExchangeServiceBase implements Closeable {
       this.credentials = null;
     }
 
-    // Reset the httpContext, to remove any existing authentication cookies from subsequent request
-    initializeHttpContext();
+    // Reset the cookies, to remove any existing authentication cookies from subsequent request
+    cookieStore.clear();
   }
   
   /**
@@ -820,6 +805,48 @@ public abstract class ExchangeServiceBase implements Closeable {
   public Map<String, String> getHttpHeaders() {
     return this.httpHeaders;
   }
+  
+  protected final CloseableHttpClient getHttpClient(){
+    CloseableHttpClient ret = httpClient;
+	if(ret == null){
+		synchronized(httpClientLock){
+			ret = httpClient;
+			if(ret == null){
+				Registry<ConnectionSocketFactory> registry = createConnectionSocketFactoryRegistry();
+				HttpClientConnectionManager httpConnectionManager = new BasicHttpClientConnectionManager(registry);
+				AuthenticationStrategy authStrategy = new CookieProcessingTargetAuthenticationStrategy();
+			
+				httpClient = ret = HttpClients.custom()
+				  .setConnectionManager(httpConnectionManager)
+				  .setTargetAuthenticationStrategy(authStrategy)
+				  .build();
+			}
+		}
+	}
+	return ret;
+  }
+  
+  protected final CloseableHttpClient getHttpPoolingClient(){
+    CloseableHttpClient ret = httpClient;
+	if(ret == null){
+		synchronized(httpPoolingClientLock){
+			ret = httpPoolingClient;
+			if(ret == null){
+				Registry<ConnectionSocketFactory> registry = createConnectionSocketFactoryRegistry();
+				PoolingHttpClientConnectionManager poolingHttpConnectionManager = new PoolingHttpClientConnectionManager(registry);
+				poolingHttpConnectionManager.setMaxTotal(maximumPoolingConnections);
+				poolingHttpConnectionManager.setDefaultMaxPerRoute(maximumPoolingConnections);
+				AuthenticationStrategy authStrategy = new CookieProcessingTargetAuthenticationStrategy();
+			
+				httpPoolingClient = ret = HttpClients.custom()
+				  .setConnectionManager(poolingHttpConnectionManager)
+				  .setTargetAuthenticationStrategy(authStrategy)
+				  .build();
+			}
+		}
+	}
+	return ret;
+  }
 
   // Events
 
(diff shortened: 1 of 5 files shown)


azurecla:
Hi __@candrews__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!

TTYL, AZPRBOT;

src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
```diff
@@ -132,24 +134,26 @@
...
-
-  private Map<String, String> httpResponseHeaders = new HashMap<String, String>();
+  private Map<String, String> httpHeaders = new ConcurrentHashMap<String, String>();
 
   private WebProxy webProxy;
 
-  protected CloseableHttpClient httpClient;
+  private volatile CloseableHttpClient httpClient;
+
+  private final Object httpClientLock = new Object();
```
serious6: why you didnt use [ReentrantLocks](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/ReentrantLock.html)?

serious6:
@candrews thanks for your contribution. I think this needs a little more review. Any chance of adding some tests to prove thread-safety?

src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
candrews: Because synchronized keyword locking is reentrant. If we wanted to avoid use of the synchronized keyword and implement the locking manually then ReentrantLock, as you said, would be the way to go... But I don't see the need for that complexity (the keyword approach is more readable).

candrews:
What do you envision such a test to look like / how would you like it to work? (I don't think there is a way to prove thread safety in unit or integration tests - we can do some concurrency testing but it wouldn't be a proof).

BTW, the double check locking implementation is textbook - there's no creativity/originality there so I don't believe it to be risky or flawed in any way. For the code that this essentially is a copy of, see https://en.m.wikipedia.org/wiki/Double-checked_locking#Usage_in_Java

I'm very eager to get this fix in as the library is more or less unusable without it in any meaningful way, so please help me in getting this merged :-)

Also, it's unfortunate that this locking in necessary at all. If ExchangeService was refactored to be immutable and have its instances created using a builder pattern, then no locking would be necessary improving runtime performance and simplicity beyond what the current design allows. Would you be open to such an API change?

serious6:
As this api is a simple portation form c# to java I am completely open to anything which improves the design and usability of it. We also had lots of PR that where already targeting these kind of issues. Nevertheless I know that there are plenty of places that are far away from good design. 
If you like to provide any contribution in that area you are welcome and I think others will support too.

src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
```diff
@@ -888,15 +893,14 @@ private void saveHttpResponseHeaders(Map<String, String> headers) {
...
-      if (ExchangeServiceBase.binarySecret == null) {
-        Random randomNumberGenerator = new Random();
-        ExchangeServiceBase.binarySecret = new byte[256 / 8];
-        randomNumberGenerator.nextBytes(binarySecret);
-      }
-
-      return ExchangeServiceBase.binarySecret;
-    }
+	Random randomNumberGenerator = new SecureRandom();
+	final byte[] bytes = new byte[256 / 8];
```
serious6: while refactoring please avoid magic numbers.

candrews: As you noted, I only refactored this magic number containing code; I don't know the significance of the magic numbers (256 and 8). Do you know or do you know how I can find out so I make them into appropriately named constants?

I see these lines were added in https://github.com/OfficeDev/ews-java-api/blame/45e3fe6301e5824a1186e9e279200df5d57d73fe/src/main/java/microsoft/exchange/webservices/data/ExchangeServiceBase.java but the commit comment ("Initial Commit") isn't helpful, and I don't know where else to look.

candrews:
I'll look into implementing a builder style ExchangeService - thanks for the info. In the mean time, is this something you could merge?

src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
vboctor: Since this is an existing issue and is unrelated, we should keep it as is.  If you desire to fix it, then do it as a separate PR.  Let's keep PRs focused on one issue.

src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
```diff
@@ -230,6 +208,9 @@ private void initializeHttpPoolingClient() {
   public void setMaximumPoolingConnections(int maximumPoolingConnections) {
     if (maximumPoolingConnections < 1)
       throw new IllegalArgumentException("maximumPoolingConnections must be 1 or greater");
+    if(httpPoolingClient!=null){
```
vboctor: Shouldn't we have space after the "if" and around the != and after the ")"?  "if (httpPoolingClient != null) {"

src/main/java/microsoft/exchange/webservices/data/core/WebProxy.java
```diff
@@ -45,8 +45,7 @@
    * @param port proxy port.
    */
   public WebProxy(String host, int port) {
-    this.host = host;
-    this.port = port;
+    this(host,port,null);
```
vboctor: Include a space after commas

src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
```diff
@@ -132,24 +134,26 @@
...
 
-  private Map<String, String> httpHeaders = new HashMap<String, String>();
-
-  private Map<String, String> httpResponseHeaders = new HashMap<String, String>();
+  private Map<String, String> httpHeaders = new ConcurrentHashMap<String, String>();
 
   private WebProxy webProxy;
 
-  protected CloseableHttpClient httpClient;
+  private volatile CloseableHttpClient httpClient;
```
vboctor: Do we really need volatile here?  Doesn't the use of synchronized between the two if checks does the trick of creating a barrier that makes the compiler assume that non local variables may be changed by another thread?

src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
```diff
@@ -132,24 +134,26 @@
...
 
   /**
    * The server info.
    */
   private ExchangeServerInfo serverInfo;
 
-  private Map<String, String> httpHeaders = new HashMap<String, String>();
-
-  private Map<String, String> httpResponseHeaders = new HashMap<String, String>();
+  private Map<String, String> httpHeaders = new ConcurrentHashMap<String, String>();
```
vboctor: Why do we need concurrent hash map for instance variables on ExchangeService?  Are we designing with the assumption that:
- A single instance of ExchangeService can be used by a single server.
- Multiple threads can run in parallel each using its own instance of ExchangeService.

src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
```diff
@@ -252,29 +233,37 @@ public void setMaximumPoolingConnections(int maximumPoolingConnections) {
...
-      } catch (IOException e) {
-        LOG.debug(e);
+      synchronized(httpPoolingClientLock){
+    	if (httpPoolingClient != null) {
+          try {
+            httpPoolingClient.close();
+          } catch (IOException e) {
+            LOG.debug(e);
+          }
+          httpPoolingClient=null;
```
vboctor: Add spaces around the equal sign.

src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
```diff
@@ -383,8 +370,6 @@ private void prepareHttpWebRequestForUrl(URI url, boolean acceptGzipEncoding, bo
     prepareCredentials(request);
 
     request.prepareConnection();
-
-    httpResponseHeaders.clear();
```
vboctor: Why is this no longer needed?

src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
candrews: I believe we do need to use ConcurrentHashMap here. I believe that ExchangeService should represent an Exchange server, and as such, it can be used by multiple threads concurrently.

The API currently allows the user to change the httpHeaders (IMHO, this variable would be better named "httpRequestHeaders" for clarity) at any time. In other words, the httpHeaders could have entries added or removed while another thread is also adding or removing entries while another thread is actually doing requests/responses. For that reason, httpHeaders must be thread safe. In my opinion, the API right now is definitely not ideal - that's why I recommended the API be changed to use an immutable ExchangeService with a builder approach, that way, the request headers would be provided using the builder and thereafter be immutable, eliminating this complexity - but we should handle that in a separate pull request.

If a user wants to connect to multiple Exchange servers, then they'll need one instance of ExchangeService for each server.

A user could create multiple ExchangeService instances for the same Exchange server, but that's not necessary, is less efficient, and more complex to write.

So, directly replying to each point:
A single instance of ExchangeService can be used by a single server: Yes.
Multiple threads can run in parallel each using its own instance of ExchangeService: Yes
Static methods are thread safe: Yes

src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
candrews: Yes, volatile is absolutely required here in order to ensure coherency between multiple threads. Please see https://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_Java for a full explanation.

src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
candrews: Okay, we'll handle the changing of "256" and "8" into constants in another issue/PR. Sounds good!

src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
```diff
@@ -252,29 +233,37 @@ public void setMaximumPoolingConnections(int maximumPoolingConnections) {
...
 
   @Override
   public void close() {
-    try {
-      httpClient.close();
-    } catch (IOException e) {
-      LOG.debug(e);
+    if(httpClient!=null) {
+	  synchronized(httpClientLock){
+	    if(httpClient!=null) {
```
vbauer: Maybe it will be better to move this code in something like IOUtils.closeQuietly to prevent copy-paste (and cleanup other places)?

src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
```diff
@@ -820,6 +805,48 @@ public void setWebProxy(WebProxy value) {
   public Map<String, String> getHttpHeaders() {
     return this.httpHeaders;
   }
+  
+  protected final CloseableHttpClient getHttpClient(){
```
vbauer: How to configure/use custom httpClient or poolingHttpClient (with some additional configuration)? Maybe it will be better to move some part of code in factory methods?

vbauer:
Just a few thoughts:

* I think it would be great to have some tests to check thread-safety, otherwise it's just a guessing.
* I'm not sure that it's a good idea to design code around "synchronized" blocks if it is not necessary (but I understand that it could be a possible solution), because of theirs bottleneck-nature. Maybe it is possible to re-write code with object pool or use only prototype objects.

src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
candrews: Sure, I can do that. But wouldn't it be better to do that in another pull request?

I'm happy to do it here, but to me, it seems like this pull request should solve one issue (thread safety) and not expand to other issues (like this kind of refactoring/cleanup). Let me know.

src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
candrews: Sure, I can do that. But wouldn't it be better to do that in another pull request?

I'm happy to do it here, but to me, it seems like this pull request should solve one issue (thread safety) and not expand to other issues (like this one, which has already been reported separately as #423). Let me know.

candrews:
@vbauer I don't think it's guessing - to prove thread safety, you have to do mathematical proofs. Thread safety isn't provable using unit tests. If you search the web for this thread safety testing, you'll find that this is the case - for example, http://programmers.stackexchange.com/questions/149374/testing-thread-safety explains the situation a bit.

I could add some tests, but I really don't want to get bogged down adding tests that really don't prove anything and IMHO don't add any value. I think that most people who are familiar with multithreaded programming would agree that this code is thread safe - perhaps you could ask someone familiar with multithreaded programming whom you trust to review this code instead?

Furthermore, I'm trying to work with you and get my contribution accepted for my own benefit and everyone else's - if there's some significant amount of other work you want done, please feel free to help as well - I'll do whatever I reasonably can, but I'm not going to spend an exorbitant amount of time on this issue. I already have deadlines that I'm in threat of missing so I will have to abandon this PR and fork the library at some point, I'd just very much rather not do that. So please help me help all of us by providing some clarity and help.

Regarding synchronized blocks, I completely agree with you that it's not a good idea to design code around unnecessary synchronized blocks. However, there are no such instances of unnecessary synchronized blocks here. There are 4 uses of synchronized: 2 to get the http clients (one each for pooling and non-pooling) and 2 to close the http clients (again, one each for pooling and non-pooling). In both cases, double checked locking is used so that the synchronized keyword is only hit (and therefore the lock is only acquired) only when absolutely necessary - under normal operation, there is zero locking. If you have a suggestion as to how to improve this design further (I already suggested a builder/immutable pattern, but such an API change is beyond the scope of this work), please share and if it's a reasonable effort I'll definitely implement it.

This thread safety issue is *HUGE* - the library is currently useless. Can you please clearly tell me what I need to do to get this PR accepted? So far, my understanding of the status is:
* I have to adjust some white space
* I'm not sure what you want done about the "magic number" situation with "256" and "8" - it seems beyond the scope of this PR. Can you please clarify?
* Are you happy with the ConcurrentHashMap? Or do I need to do anything for that?
* Refactor to use a common "closeQuietly" method (why is the scope of this work expanding?)
* Resolve issue #423 (again, why is the scope of the issue expanding?)

candrews:
Any news?

Thanks! I hope you enjoy the long weekend

src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
serious6: requirements are defined in [Googles Java Guide](https://google.github.io/styleguide/javaguide.html) which has been discussed to be our project reference. This is also as well mentioned in the [CONTRIBUTING.md](https://github.com/OfficeDev/ews-java-api/blob/master/CONTRIBUTING.md#coding-conventions). Using checkstyle has as well been discussed but comes along with several problems:
* usability of checkstyle-plugins (for different IDEs) differ from the checkstyle-google-styleguide
* project contains multiple formattings that where not touched by IntelliJ for some reasons these would have to be handled manually, maybe there are also bugs in the `codeStyleSettings.xml` didnt had that much time to step in.
If you have any ideas on how to do this "at low cost" feel free to discuss at #4. You are pushing at open doors with this. But whether or not we have these automated checks every PR has to match these criterias.

serious6:
to answer your questions:
1. Matching the styleguide is necessary yes. Sorry that we dont have central checkstyle configuration but if needed there is also the possibility to run local checks. Modifying the suggested lines shouldnt be that much of work btw.
2. is already answered by @vboctor 
3. using `ConcurrentHashMap` should be ok at this point. Any follow up PR relating this topic are welcome. 
4. +1 for another PR
5. +1 for another PR

Some sidenodes:
> I think that most people who are familiar with multithreaded programming would agree that this code is thread safe - perhaps you could ask someone familiar with multithreaded programming whom you trust to review this code instead?

There is no doubt that you will have tested your code. In general the reason to add unit-tests is to guarantee quality also over the next changes that will possibly be done by others. Regarding your point of a 'trusted person' we so far put our PRs under review by the community so everyone is able to comment. And as you might have noticed there is some related feedback improving the quality of this PR. To solve this I would suggest to include your fix in the next version. So we will have this merged after 2.0 has been released. So maybe we will be able to also include some follow up PR on this one. Any comments?

src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
vboctor: I had a typo that was critical here:
"A single instance of ExchangeService can be used by a single server."  I meant a single instance is used by a single thread.  i.e. ExchangeService is not thread safe, but it is OK to have 3 threads using 3 different instances of ExchangeService connecting to same or difference servers in parallel.

It seems that you want to have a single instance of ExchangeService used by multiple threads in parallel.  What is the motivation for that.  That adds complexity and will have weird cases if the state can depend on the request (e.g. if headers may need to change based on request for example).

> I believe we do need to use ConcurrentHashMap here. I believe that ExchangeService should represent
> an Exchange server, and as such, it can be used by multiple threads concurrently.

If we go with the model that I'm suggesting, then that we don't need this protection against concurrency.

vboctor:
I'm still not sold on the need for this PR, so I added added comment with my question and explaining my point.  I may be convinced though.

src/main/java/microsoft/exchange/webservices/data/core/ExchangeServiceBase.java
candrews: The way ExchangeService is implemented today it isn't thread safe even when you create new instances of it for each thread - see #371.

If it is thread safe (as this PR makes it), ExchangeService becomes:
# Easier to use. I believe developers would assume it's thread safe, as *Service named classes usually are. It's also less code to use it thread safe rather than creating, configuration, then closing it for each use.
# Faster. Before this PR, the suggested proper use of ExchangeService is to create and then tears down an HttpClient every time it's used - that takes time. It's also unable to share HttpClient resources (like thread pools) as the HttpClient design intends, reducing performance.
# Less memory intensive. By having fewer instances of ExchangeService, less memory is used. Most significantly, by sharing HttpClient instances, thread pools, caches, and objects are all shared (as HttpClient is intended to be used).

candrews:
1. You're right in that the whitespace changes aren't a big deal. I've just added a commit (04a3200806429057297f8c1c379c4cfe05411f36) addressing them and ensured that checkstyle passes now. If you'd like, I could rebase/merge this PR's now 2 commits into one - just let me know.
In order to prevent myself (and others!) from causing whitespace headaches for you going forward, I submitted a PR adding checkstyle to the build: #438
2. :+1: 
3. :+1: 
4. I submitted PR for that: #439 
5. Managing the scope of a shared HttpClient between ExchangeService instances seems like it'll be tricky: the current API hides the httpClient from the user, including closing it when ExchangeService is closed. If we allow the user to specify an httpClient, then should ExchangeService close it when it's done or not? And if the user doesn't specify one, should ExchangeService close it? I don't think it's insurmountable, but I'll leave that for later or for someone else to tackle. Ideally, I think a builder for ExchangeService with an immutable ExchangeService API should be created, and the specifiable httpClient feature can be added at the same time.

I also replied to @vboctor's comment/concerns at https://github.com/OfficeDev/ews-java-api/pull/428#discussion_r38964148

Thanks again

codewheeney:
Are there any updates for this PR?  We're actively blocked by this issue at the moment.

codewheeney:
I cloned off of candrews fix and tested, and this solution is necessary but not sufficient to fix the concurrency issue I'm seeing.  There appears to be a bug in Apache's CloseableHttpClient, namely that, although it's marked as thread safe, it uses the not threadsafe NTLMScheme class.  Concurrent access can cause out of bounds exceptions when writing to internal buffers due to static objects in NTLMEngineImpl. 

I've opened this issue with apache:  https://issues.apache.org/jira/browse/HTTPCLIENT-1686

codewheeney:
The issue I reported about HttpClient has been closed as invalid, although I think the maintainer of that code doesn't quite see my point, unless I'm not following what this PR is attempting to do.  Fundamentally, this PR does the right thing by creating a separate HttpClientContext for each concurrent access.  That should be sufficient according to the comments in HTTPCLIENT-1686, but it isn't in my testing because of the static field in NTLMEngineImpl.java.

I'd appreciate any other insight into testing or use of this code in this PR that others are seeing.

schakko:
@candrews Thank you for this pull request, but it doesn't solve all concurrency problems: I am stumbling upon the exactly same problem @codewheeney described.

In our case we have two threads (one HTTP worker thread, one background thread) which are accessing the Exchange web service through multiple instances of ExchangeService. 
HTTP worker thread pulls randomly data from the Exchange web service, the background check pulls data every two minutes. If the requests are occuring at the same time, the exception

	Caused by: java.lang.ArrayIndexOutOfBoundsException: 40
		at org.apache.http.impl.auth.NTLMEngineImpl$NTLMMessage.addByte(NTLMEngineImpl.java:911)
		at org.apache.http.impl.auth.NTLMEngineImpl$NTLMMessage.addULong(NTLMEngineImpl.java:940)
		at org.apache.http.impl.auth.NTLMEngineImpl$Type1Message.getResponse(NTLMEngineImpl.java:1043)

occurs. It doesn't matter if I use the same instance of ExchangeService, create multiple instances for both threads or set the maximumPoolingConnections to 1 in any of those combinations.

At the moment, @codewheeney suggestion to modify the HttpClient seems to be the only working solution - which I do want to avoid b/c of obvious reasons.
Are there any other ideas or did you made progress in this case, @codewheeney ?

codewheeney:
I have not re-opened https://issues.apache.org/jira/browse/HTTPCLIENT-1686 yet.  At the moment, I've forked HTTP Client and EWS into our repository and I'm working from patched versions. 

It'd be great if @schakko commented on https://issues.apache.org/jira/browse/HTTPCLIENT-1686 and re-opened it, pointing out that the static data member in NTLMEngineImpl will never work properly in a multithreaded environment.  I failed to get that point across and haven't had cycles to get in there and re-argue the case.

schakko:
@codewheeney I left a comment on issues.apache.org but it's hard to argue if they don't want to make it thread-safe (and more reliable).
BTW: I fixed the EWS API without touching the HTTP client. Feel free to pull my branch from https://github.com/schakko/ews-java-api/tree/gh-371-make-ews-thread-safe.

wwahmed:
Hello @schakko 

I see the changes you have made to make the API more threadsafe. However, I am confused about your comment that you have done so without touching the HTTPClient when the fork you are on, had the commit #5541396 from lamielle which indicates that it's being built with the forked httpclient (httpclient-netc, upgrade to 4.5.1) and it's still present in pom.xml in your commit. 

Do the changes in your fork require everything done previously by candrews and lamielle - including the httpclient update?

Thanks for your help.

schakko:
@wwahmed I understand your confusion.
I compiled and (unit-)tested it against 4.4.1 and it did the job. But I must admit that I did not test it in production.

Our working ews-java-api has the following history:
- 690c75e: was still in our local repository, pushed it a few seconds ago; does only change httpclient-netc to httpclient and removes the custom repository. Revokes 55413968488
- 1903b73aea2: my fix, does compile/test against 4.4.1
- 7cca414b32b: obsolete; no code changes
- 55413968488: introduces httpclient-netc-4.5.1, revoked through 690c75e
- 66b107b9c25: coding standards
- 0020b636eb0: initial work for this issue against httpclient 4.4.1

The build with this history has been running in production for 3 months without any problems. In this time the API has been accessed by ~8-10 threads concurrently all the time.

wwahmed:
Hi @schakko  Thanks for confirming - I have also tested it without the httpclient-netc-4.5.1 http client and it does work with 8-10 concurrent threads. 

Thanks a lot for your confirmation.

serious6:
@schakko maybe you can share the unit tests you created? Or provide a PR in this? Increasing the coverage might ve useful.

SebastianWe:
When using multiple threads a "ServiceRequestException: The request failed. Connection is still allocated" might occure.

I added a fix for this in #491.

PR #439, opened 2015-09-08
Use IOUtils.closeQuietly instead of reimplementing that method in each location

candrews:
Add a dependency on commons-io and use it appropriately.

Unfortunately, javax.xml.stream.XMLStreamWriter doesn't implement java.io.Closeable, so IOUtils.closeQuietly couldn't be used in a few places.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,6 +80,7 @@
         <commons-logging.version>1.2</commons-logging.version>
         <joda-time.version>2.8</joda-time.version>
         <commons-lang3.version>3.4</commons-lang3.version>
+        <commons-io.version>2.4</commons-io.version>
         <!--  Dependencies [TEST]:  -->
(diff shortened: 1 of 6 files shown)


azurecla:
Hi __@candrews__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!

TTYL, AZPRBOT;

pom.xml
```diff
@@ -176,6 +177,12 @@
         </dependency>
 
         <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
+            <version>${commons-io.version}</version>
```
serious6: do we really need another dependency for this? Sources of `closeQuietly(...)` is really primitive

serious6:
@candrews merged. Thanks for your contribution.

Issue #425, opened 2015-08-25
AlternateMailboxCollection parses only the first entry


thomaswoeckinger:
The loadFromXML(...) method of AlternateMailboxCollection checks the wrong end element, therefore only the first AlternateMailbox entry will be parsed and returned.

XmlElementNames.AlternateMailbox must be replaced with XmlElementNames.AlternateMailboxes

The code should be corrected to:

```
  public static AlternateMailboxCollection loadFromXml(EwsXmlReader reader)
      throws Exception {
    AlternateMailboxCollection instance = new AlternateMailboxCollection();

    do {
      reader.read();

      if ((reader.getNodeType().getNodeType() == XmlNodeType.START_ELEMENT) &&
          (reader.getLocalName()
              .equals(XmlElementNames.AlternateMailbox))) {
        instance.getEntries().add(
            AlternateMailbox.loadFromXml(reader));
      }
    } while (!reader.isEndElement(XmlNamespace.Autodiscover,
        XmlElementNames.AlternateMailboxes));

    return instance;
  }
```

serious6:
thanks. Please provide a PR with your suggested code. Any related tests on this issue are welcome.

thomaswoeckinger:
The code is already provided in the original post, but if it is necessary i can post a patch.

thomaswoeckinger:
diff --git a/src/main/java/microsoft/exchange/webservices/data/autodiscover/AlternateMailboxCollection.java b/src/main/java/microsoft/exchange/webservices/data/autodiscover/AlternateMailboxCollection.java
index 5e72944..74e48a9 100644
--- a/src/main/java/microsoft/exchange/webservices/data/autodiscover/AlternateMailboxCollection.java
+++ b/src/main/java/microsoft/exchange/webservices/data/autodiscover/AlternateMailboxCollection.java
@@ -23,13 +23,13 @@
 
 package microsoft.exchange.webservices.data.autodiscover;
 
+import java.util.ArrayList;
+import java.util.List;
+
 import microsoft.exchange.webservices.data.core.EwsXmlReader;
 import microsoft.exchange.webservices.data.core.XmlElementNames;
 import microsoft.exchange.webservices.data.core.enumeration.misc.XmlNamespace;
 import microsoft.exchange.webservices.data.security.XmlNodeType;
-
-import java.util.ArrayList;
-import java.util.List;
 
 /**
  * Represents a user setting that is a collection of alternate mailboxes.
[...]

PR #496, opened 2016-02-15
Issue #425: Fix hanging StreamingSubscriptionConnection.close()

dconnelly:
StreamingSubscriptionConnection.close() calls HttpWebRequest.close() which will first
attempts to consume the response before releasing the connection. This unfortunately
will hang since the notification reader thread is still blocked reading from the same 
connection. The fix is to just the connection without consuming the response entity.

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/core/request/HangingServiceRequestBase.java
+++ b/src/main/java/microsoft/exchange/webservices/data/core/request/HangingServiceRequestBase.java
@@ -266,7 +266,7 @@ public abstract class HangingServiceRequestBase<T> extends ServiceRequestBase<T>
    */
   public void disconnect() {
     synchronized (this) {
-      IOUtils.closeQuietly(this.response);
+      this.response.releaseConnection();
       this.disconnect(HangingRequestDisconnectReason.UserInitiated, null);
     }
   }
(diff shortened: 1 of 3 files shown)


azurecla:
Hi __@dconnelly__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!

TTYL, AZPRBOT;

src/main/java/microsoft/exchange/webservices/data/core/request/HangingServiceRequestBase.java
```diff
@@ -239,13 +238,26 @@ private void parseResponses() {
...
+        // Stream was closed due to an error
+        this.disconnect(HangingRequestDisconnectReason.Exception, ex);
+      }
+
     } finally {
       IOUtils.closeQuietly(responseCopy);
     }
   }
 
+  private static boolean isSocketClosed(Throwable e) {
```
serious6: Is this about getting smth like the [rootcause](https://commons.apache.org/proper/commons-lang/javadocs/api-2.4/org/apache/commons/lang/exception/ExceptionUtils.html#getRootCause(java.lang.Throwable))?

Issue #497, opened 2016-02-15
Not receiving heartbeats with StreamingSubscriptionConnection


dconnelly:
The code seems to expect a heartbeat response every 45s but I'm not seeing any, at least while connected to Office365. This causes the underlying http connection to time out before lifetime minutes is reached, unless I've made sure that ExchangeService has a much larger timeout than would be applicable to other requests.

A simple fix would be to set the http request timeout to lifetime minutes just for the hanging GetStreamingEvents call.

PR #499, opened 2016-02-15
Issue #497: Allow socket timeout to be overridden on a per request basis

dconnelly:
This allows the GetStreamingEvents call to specify a longer timeout so connection is not
dropped before lifetime minutes has been reached. Relying on heartbeats to keep the
connection alive does not seem to work for Office365.

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/core/request/ServiceRequestBase.java
+++ b/src/main/java/microsoft/exchange/webservices/data/core/request/ServiceRequestBase.java
@@ -72,6 +72,9 @@ public abstract class ServiceRequestBase<T> {
    */
   private ExchangeService service;
 
+  /* If included then overrides service specified timeout */
+  private Integer timeout;
+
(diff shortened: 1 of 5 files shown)


azurecla:
Hi __@dconnelly__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!

TTYL, AZPRBOT;

src/main/java/microsoft/exchange/webservices/data/core/request/ServiceRequestBase.java
```diff
@@ -122,6 +125,10 @@ protected void validate() throws Exception {
     this.service.validate();
   }
 
+  public void setTimeout(Integer timeout) {
```
serious6: please add javadoc to this method

serious6:
once javadocs has been added this is ready to merge

PR #501, opened 2016-02-20
Remove unstable part in Readme

WtfJoke:
Remove part of readme, that indicates, there isnt any stable version out yet.

Since, there is a version 2.0 out, there isnt any need for this information.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -6,7 +6,7 @@
 Please see the [Getting Started Guide](https://github.com/OfficeDev/ews-java-api/wiki/Getting-Started-Guide) on our wiki for an introduction to this library.
 
 ## Using the library
-Prebuilt JARs are available in the Maven Central repository, which are easy to use with your project. See next section for further details.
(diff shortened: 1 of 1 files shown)


serious6:
Maybe we should just link to the wiki pages and update the info there?

WtfJoke:
Makes sense. I'll update the pr this noon.
I'll also fix the entry in the wiki, because there is the same wrong text as well.

WtfJoke:
@serious6 I'll updated the readme to the link. 

There are some duplicate informations on the wiki.
For example the maven/gradle part can be either found here:
https://github.com/OfficeDev/ews-java-api/wiki#maven--gradle-integration
or here:
https://github.com/OfficeDev/ews-java-api/wiki/Getting-Started-Guide#maven--gradle-integration

Shouldnt we delete informations on the first page? So we dont need to update two pages.

serious6:
Thanks for pointing this out. The information should be centralized while updating the remaining references. Maybe the sections on the readme should also be more generalized/combined while pointing to the wiki pages.

serious6:
@WtfJoke thanks for your contribution

PR #454, opened 2015-11-08
update dependencies to latest version

serious6:

PR #536, opened 2016-06-16
Removed IntelliJ files

vanta:
Related to #454


azurecla:
Hi __@vanta__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

serious6:
@vanta the reason why we made the integration of the idea specific project files is that we wanted people to be enforced to use the documented formatter. From my point of view just removing the files does not resolve the conflict. Any comments?

vanta:
@serious6 Thanks for your comment. I've got your point. I'm an IntelliJ user as well so glad to hear that but how about other IDEs' users? Formatting still can be messed up as the is no way to enforce formatting in IDE (user can always change it or don't format).

Maybe this formatting check can be one of build steps (by using some plugin)?

serious6:
@vanta thanks for your support.

Issue #579, opened 2016-11-25
Authenticate and fetch EWS data using OAuth token


Surbhidhingra:
I am trying to build an android app using **EWS-java-api** to fetch EWS data with an oAuth token using[ active directory authentication lib](https://github.com/AzureAD/azure-activedirectory-library-for-js) for android. similar to what is shown in this article for[ .net](https://msdn.microsoft.com/en-us/library/office/dn903761(v=exchg.150).aspx)

Below is the code: 

// Code to acquire token after registering the native application in Azure active directory

`authenticationContext.acquireToken(<activity context>, 
"<resource id: copied from the manifest file tag <resourceAppId> of Azure active directory after adding permission>",
"<Application id of the registered app in AAD>", 
"<Application Redirect URI>",
email, PromptBehavior.Always, "",
AuthenticationCallback);`

//We receive **AuthenticationResult** object containing authentication token in **AuthenticationCallback** onSuccess method and then call an async task to fetch EWS data
 
`ExchangeService exchangeService = new ExchangeService();
exchangeService.setTraceEnabled(true);
exchangeService.getHttpHeaders().put("Authorization", "Bearer " + mAuthenticationResult.getAccessToken());
exchangeService.setUrl(URI.create("https://outlook.office365.com/EWS/Exchange.asmx"));`

I am able to get oAuth token, but i am not able to receive exchange data, it gives me unauthorised and forbidden access when fetching data using below code.

` CalendarFolder cf = CalendarFolder.bind(service, WellKnownFolderName.Calendar);
findResults = cf.findAppointments(new CalendarView(startDate, endDate));`

Also, i am not sure about the configuration i am setting up in azure portal. It would be great if you can tell how to setup Azure AD application for fetching EWS data via oAuth authentication in android.

kb000:
From debugging and reading the code, it looks like there's no plumbing for the Bearer token (aka access_token) to be added to the http request.

Looks like a good candidate for a pull request.

Disclaimer: I am a Microsoft Employee, but am using this framework as a student.

dsebastien:
Any news about this?
Even with the class you've developed, we always end up with an authorization error, although the token was just retrieved and is still valid.

I'm not familiar with the code so it's a bit unclear to me what the EmailMessage.bind(...) actually does (i.e., call a REST or SOAP API). Don't know if it might help, but my scenario is as follows:

[...]

PR #606, opened 2017-05-02
Add microsoft.exchange.webservices.data.credential.BearerTokenCredentials

kb000:
This should address https://github.com/OfficeDev/ews-java-api/issues/579

### DIFF
--- a/src/main/java/microsoft/exchange/webservices/data/credential/BearerTokenCredentials.java
+++ b/src/main/java/microsoft/exchange/webservices/data/credential/BearerTokenCredentials.java
@@ -1,0 +1,90 @@
+/*
+ * The MIT License Copyright (c) 2017 Kevin Burek <khb718@g.harvard.edu>
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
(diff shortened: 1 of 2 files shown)


dsebastien:
Hello,

Any chance to get this PR merged in the near future?
Am I mistaken if I think that without this, using the https://github.com/bookaroom/ews-java-api/blob/master/src/main/java/microsoft/exchange/webservices/data/credential/TokenCredentials.java is useless and doesn't work?

Currently I try to bind to an email item, but always end up with a 403 unauthorized error...

kb000:
From what I understand, TokenCredentials.java was designed to be used with a SAML or Kerberos tokens.  The heavy lifting is done in WSSecurityBasedCredentials, the base class.  I don't know much more about it than that it didn't work for me, living in the OAuth world of 2017.

I found success using AAD to get an OAUTH2 bearer token (e.g. from a POST to https://login.windows.net/common/oauth2/token?api-version=1.0).  This PR allows the ews-java-api to pass that token along to the EWS SOAP service. 

I was inspired by this StackOverflow answer, where someone was able to use EWS/SOAP directly with a bearer token:  https://stackoverflow.com/questions/41875795/office-365-calendar-api-using-ews-and-oauth-2

PR #609, opened 2017-05-17
Various improvements and bug fixes

beders:
Added AutodiscoverUnauthorizedException which will be thrown when auto-discovery requires credentials and given credentials are not valid.

CHanged XML parser to ignore invalid entities that can appear for unknown reasons.

Fixed method signature in loadPropertiesForItems so Calendar items can be loaded in batch as well.

Added missing time zone entries.


beders:
PR is not according to contribution guidelines

PR #620, opened 2017-06-07
Remove joda and switch to use java.time (requires java 8)

robneild:


robneild:
Obviously a bit contentious. Also, I'm not 100% sure about some of the time zone stuff. Anyway here is the code if anyone wants it

PR #409, opened 2015-08-17
Force XML v1.1 preable to avoid parse errors - WIP

easel:
Relates to #353. This is a work in progress patch. It currently introduces a dependency on stream flyer and does not have meaningful tests. It's been a while since I created it, so if anybody has a failing case to test against it would be helpful.

PR #727, opened 2020-04-06
SOC-555: Force XML v1.1 preable to avoid parse errors.

aglumova:
Sometimes you have to open XML documents that contain characters that are allowed in XML 1.1 documents but not allowed in XML 1.0 documents. And sometimes you have to open XML documents that contain characters that are entirely forbidden. For these kind of documents some pre-defined modifier exist so that the modified stream can be opened by standard XML parsers:

---

Apply fix from https://github.com/OfficeDev/ews-java-api/pull/409/files


girishghoda57:
Hi @aglumova this PR is solving for all the invalid characters issue or for only attachment related issues?
We applied changes but we get this error "microsoft.exchange.webservices.data.core.exception.service.remote.ServiceRequestException: The request failed. ParseError at [row,col]:[1,20]
Message: A pseudo attribute name is expected. "
Any help on this? Thanks in advance

PR #733, opened 2020-08-11
Email not ingest due to control character in email.

girishghoda57:
## Issue
Email not ingest due to control character in email.
"Caused by: javax.xml.stream.XMLStreamException: ParseError at Message: Character reference ""&# ".


zhangyinghahaha:
Thx very much!

